PR #178, opened 2026-05-23
feat: parse and validate optional mvp_scope in goal contract

KyubumShin:
## What

Stage A implementation step 1: introduce `mvp_scope` as an optional top-level field on `.mpl/goal-contract.yaml`. Parser + validator + schema doc only — no orchestrator behavior change yet.

- `hooks/lib/mpl-goal-contract.mjs` (+39): `parseMvpScope` helper, `mvp_scope` field on parsed contract, validation rules (id-existence + artifact enum + non-empty-when-present), exported `MVP_SCOPE_ARTIFACTS` constant.
- `hooks/__tests__/mpl-goal-contract.test.mjs` (+82): 10 new tests covering absent / present, inline-list and block-list shapes, unknown AC/AX rejection, unsupported artifact rejection, missing artifact rejection, fully-empty rejection.
- `docs/schemas/goal-contract.md` (+29): new \"Optional: `mvp_scope` (Stage A)\" section.

Full hook test suite: **912/912 pass**.

## Why

The Stage A RFC (`docs/roadmap/stage-a-mvp-cuts-rfc.md`) requires goal-contract to carry a user-declared MVP subset as the entry point of the release path. This commit adds *only* the schema + parser + validator; orchestrator routing (`release-gate` / `release-finalize` states, cohort-aware `phase2-sprint`), interview guided-checklist (D-Q1), and decomposer `mvp` derivation are deliberately scoped out and will land in subsequent commits.

Backward compatibility: contracts without `mvp_scope` remain valid; the parsed field is `null` and existing readiness checks are unchanged.

## Design

- Source RFC: `docs/roadmap/stage-a-mvp-cuts-rfc.md` §4.1 (schema requirements).
- Validation rules: id existence against contract's `acceptance_criteria` / `variation_axes`, `artifact` enum `{draft_pr | branch | tag | release_manifest}`, non-empty when present.
- Schema doc: this PR adds the \"Optional: `mvp_scope` (Stage A)\" section to `docs/schemas/goal-contract.md`.

## Agent Review Status

- [ ] from claude
- [ ] from codex

---

### DIFF
--- a/docs/schemas/goal-contract.md
+++ b/docs/schemas/goal-contract.md
@@ -99,9 +99,13 @@ Validation rules (enforced by `hooks/lib/mpl-goal-contract.mjs`):
 - `mvp_scope.artifact` MUST be one of the values in `MVP_SCOPE_ARTIFACTS`
   (`draft_pr | branch | tag | release_manifest`).
 
-When `mvp_scope` is declared, the orchestrator routes MVP completion through
-the Stage A release path (`release-gate` → `release-finalize`); see
-`docs/roadmap/stage-a-mvp-cuts-rfc.md` for the full state machine.
+**Current Stage A status: schema only.** Parsing and validation are wired
+through `hooks/lib/mpl-goal-contract.mjs`; declaring `mvp_scope` today has no
+runtime effect beyond schema validation. When the Stage A release-path
+states (`release-gate` and `release-finalize`) and decomposer `mvp`
+derivation land in subsequent commits, MVP completion will route through
+the release path. See `docs/roadmap/stage-a-mvp-cuts-rfc.md` for the full
+target state machine.
 
 ## Runtime Enforcement
 
--- a/docs/schemas/goal-contract.md
+++ b/docs/schemas/goal-contract.md
@@ -74,6 +74,35 @@ deferred_uncertainties: []
 overrides: []
 ```
 
+## Optional: `mvp_scope` (Stage A)
+
+When the user declares an MVP subset during Phase 0, the contract gains an
+optional top-level `mvp_scope` field. Absent = pre-Stage-A behavior (no MVP
+cut; the current sequential pipeline runs unchanged).
+
+```yaml
+mvp_scope:                            # OPTIONAL — omit for current behavior
+  acceptance_criteria: [AC-1, AC-2]   # subset of acceptance_criteria[].id
+  variation_axes: [AX-1]              # subset of variation_axes[].id
+  artifact: draft_pr                  # draft_pr | branch | tag | release_manifest
+```
+
+Validation rules (enforced by `hooks/lib/mpl-goal-contract.mjs`):
+
+- `mvp_scope` itself is optional. Contracts without it remain valid.
+- Every `mvp_scope.acceptance_criteria[]` id MUST exist in the contract's
+  `acceptance_criteria[].id` set.
+- Every `mvp_scope.variation_axes[]` id MUST exist in the contract's
+  `variation_axes[].id` set.
+- `mvp_scope.acceptance_criteria` and `mvp_scope.variation_axes` may not
+  both be empty when `mvp_scope` is present.
+- `mvp_scope.artifact` MUST be one of the values in `MVP_SCOPE_ARTIFACTS`
+  (`draft_pr | branch | tag | release_manifest`).
+
+When `mvp_scope` is declared, the orchestrator routes MVP completion through
+the Stage A release path (`release-gate` → `release-finalize`); see
+`docs/roadmap/stage-a-mvp-cuts-rfc.md` for the full state machine.
+
 ## Runtime Enforcement
 
 - `hooks/mpl-ambiguity-gate.mjs` blocks `mpl-decomposer` dispatch when the
--- a/hooks/__tests__/mpl-goal-contract.test.mjs
+++ b/hooks/__tests__/mpl-goal-contract.test.mjs
@@ -6,6 +6,7 @@ import { tmpdir } from 'os';
 
 import {
   GOAL_CONTRACT_REL_PATH,
+  MVP_SCOPE_ARTIFACTS,
   parseGoalContractText,
   readGoalContract,
   validateGoalContractText,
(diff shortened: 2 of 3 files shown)


KyubumShin (commented):
## Review: feat — parse and validate optional `mvp_scope` in goal contract

Scope is tight and well-bounded (parser + validator + schema doc only, no orchestrator change), test coverage is strong (10 new tests, 912/912 pass). A few items worth surfacing before merge.

### Blocker

**1. Schema doc overstates what this PR implements.** `docs/schemas/goal-contract.md` lines 102–104 (added in this PR) say:

> When `mvp_scope` is declared, the orchestrator routes MVP completion through the Stage A release path (`release-gate` → `release-finalize`); see `docs/roadmap/stage-a-mvp-cuts-rfc.md` for the full state machine.

But the PR body explicitly says orchestrator routing is "deliberately scoped out and will land in subsequent commits." A user who reads only this schema doc after merge and adds `mvp_scope` to their goal-contract.yaml will reasonably expect MPL to route through `release-gate`/`release-finalize` — but no such routing exists yet. The contract is "parseable" but inert.

Two fixes either of which works:
- (a) Add a one-line "Stage A status: schema only — release-path routing lands in a follow-up commit. With this commit alone, declaring `mvp_scope` has no runtime effect." This makes the doc temporally honest.
- (b) Drop the orchestrator sentence and replace with a forward-link only: "When the Stage A release path lands, MVP completion will route through `release-gate` → `release-finalize`; see `docs/roadmap/stage-a-mvp-cuts-rfc.md`."

The doc was clearly written for the end-state. Either fix surfaces "this is not yet wired."

### Medium — silent acceptance of empty `mvp_scope:` block

**2. `parseMvpScope` returns `null` for a present-but-empty block, so the validator never sees a malformed declaration.** At `hooks/lib/mpl-goal-contract.mjs:135–143`:

```js
function parseMvpScope(text) {
  if (!/^mvp_scope\s*:/m.test(text)) return null;
  const block = extractTopBlock(text, 'mvp_scope');
  if (!block.trim()) return null;
  ...
}
```

If a user writes:

```yaml
mvp_scope:
# (nothing else under it, then another top-level key or EOF)
```

The key regex matches → `extractTopBlock` returns empty → `block.trim()` is falsy → return `null` → `validateGoalContractText` skips the `if (contract.mvp_scope)` block → contract is reported as valid with `mvp_scope: null`. Net effect: user thinks they declared MVP scope but their declaration was silently dropped. Same problem with `mvp_scope: null` or `mvp_scope: ~`.

Two ways to fix:
- (a) Treat "key present but block empty" as a malformed declaration. Return `{acceptance_criteria: [], variation_axes: [], artifact: null}` so the existing validator surfaces `mvp_scope.acceptance_criteria_or_variation_axes` and `mvp_scope.artifact`. This is the simplest signal and reuses error vocabulary already in the test suite.
- (b) Add a distinct `mvp_scope.empty` missing-key to the validator for this case.

Add a test alongside it:
```js
it('rejects mvp_scope when the block is declared but empty', () => {
  const text = validGoalContract() + 'mvp_scope:\n';
  const verdict = validateGoalContractText(text);
  assert.equal(verdict.valid, false);
});
```

### Minor — test coverage gaps

**3. No assertion that omitting `variation_axes` defaults to `[]`.** The "parses mvp_scope with block-list acceptance_criteria" test (line 175-179 in the new test block) builds a block with `acceptance_criteria` only and `artifact`, then asserts AC and artifact but not the absent-AX shape. The validator's empty-both check assumes `listAfterKey` returns `[]` for missing keys, which is true by inspection of `mpl-goal-contract.mjs:73–103`, but the test should pin this contract:

```js
it('parses mvp_scope with block-list acceptance_criteria', () => {
  const text = mvpScopeBlock('  acceptance_criteria:\n    - AC-1\n  artifact: release_manifest');
  const c = parseGoalContractText(text);
  assert.deepEqual(c.mvp_scope.acceptance_criteria, ['AC-1']);
  assert.deepEqual(c.mvp_scope.variation_axes, []);   // <— add this
  assert.equal(c.mvp_scope.artifact, 'release_manifest');
});
```

**4. No test for "only `artifact` declared, AC/AX keys entirely absent."** Functionally the same as both-empty via `listAfterKey` returning `[]`, but worth pinning since it's the most-likely user mistake (declare the artifact target but forget the scope):

```js
it('rejects mvp_scope when only artifact is declared (no AC/AX keys)', () => {
  const text = mvpScopeBlock('  artifact: branch');
  const verdict = validateGoalContractText(text);
  assert.equal(verdict.valid, false);
  assert.ok(verdict.missing.includes('mvp_scope.acceptance_criteria_or_variation_axes'));
});
```

### Nits (non-blocking)

- `MVP_SCOPE_ARTIFACTS` is `Object.freeze`'d on an Array literal. The freeze prevents reassignment of indices but doesn't make it iterable as a `Set` — consumers using `.includes()` (as the validator does at `mpl-goal-contract.mjs:256`) are fine. If you want set-like O(1) membership for future readers, `new Set([...])` would express intent more clearly. Strictly cosmetic.
- `parseMvpScope` regex `/^mvp_scope\s*:/m` will match `mvp_scope:` even if it appears in a multiline string value of another field (unlikely in practice given the prompt-controlled YAML subset documented in the file header at `mpl-goal-contract.mjs:8–14`, but worth a comment if that subset ever expands).

### Verdict

Code quality is good, the schema/test/parser triple split is clean, and the validation vocabulary (`mvp_scope.acceptance_criteria.unknown_id:AC-999`) is precise and machine-readable. Blocker is the schema doc lying about runtime behavior; the empty-block silent-null is the real correctness gap worth catching here rather than waiting for a Phase 0 user to wonder why their declaration "disappeared." Minor test coverage tweaks are quick wins.

— from claude

KyubumShin (commented):
1. **High**: `mvp_scope:`가 선언됐지만 비어 있는 계약이 유효한 “absent” 상태로 접힙니다. `parseMvpScope()`는 top-level key를 감지한 뒤에도 block이 비어 있으면 `null`을 반환하고(`hooks/lib/mpl-goal-contract.mjs:135`-`139`), validator는 `contract.mvp_scope`가 truthy일 때만 Stage A 검증을 실행합니다(`hooks/lib/mpl-goal-contract.mjs:239`). 그래서 otherwise-valid contract 뒤에 `mvp_scope:\n`만 붙이면 `validateGoalContractText()`가 `valid: true`, `contract.mvp_scope: null`을 반환합니다. 문서상 absent만 기존 동작이고(`docs/schemas/goal-contract.md:79`-`81`), present 상태에서는 AC/AX가 둘 다 비면 안 됩니다(`docs/schemas/goal-contract.md:97`-`98`). 이 상태로 Phase 0이 빈 block을 남기면 사용자는 MVP를 선언한 셈인데 후속 orchestrator/decomposer는 `mvp_scope`가 없는 것으로 보고 release path를 건너뜁니다. `mvp_scope` key presence를 별도로 보존해서 empty block은 `mvp_scope.acceptance_criteria_or_variation_axes` 및 `mvp_scope.artifact`로 reject하고, 해당 regression test를 추가하는 게 맞습니다.

확인한 테스트:
- `node --test hooks/__tests__/mpl-goal-contract.test.mjs` 통과
- `npm test` 통과 (`912 tests / 0 failures`)

— from codex

KyubumShin (commented):
## Re-review at `f789ba5` — all prior concerns addressed

Verified each item from my earlier review against the diff:

- **Blocker #1 (schema doc overstating runtime behavior)** — fixed. New "Current Stage A status: schema only" paragraph at `docs/schemas/goal-contract.md:102–108` cleanly separates "what works now" (parsing + validation) from "what lands later" (release-path routing). The forward-reference to the RFC is preserved without overpromising.
- **Medium #2 (silent acceptance of empty `mvp_scope:` block)** — fixed at `hooks/lib/mpl-goal-contract.mjs:135–145`. `parseMvpScope` no longer short-circuits on empty block; returns the empty-shape object so the validator surfaces `mvp_scope.acceptance_criteria_or_variation_axes` + `mvp_scope.artifact`. The new code comment is clear on the `null` vs `{...}` semantic split ("absent" vs "malformed"). Also handles the `mvp_scope: ~` / `mvp_scope: null` inline-null case correctly by inspection (extractTopBlock returns empty, downstream parsers default to `[]` / `null`).
- **Minor #3 (variation_axes default `[]` assertion)** — fixed. `mpl-goal-contract.test.mjs:173` now asserts `assert.deepEqual(c.mvp_scope.variation_axes, [])` in the block-list AC test.
- **Minor #4 (artifact-only, no AC/AX keys)** — fixed. New test "rejects mvp_scope when only artifact is declared (no AC/AX keys)" at `mpl-goal-contract.test.mjs:194–200`.
- **Bonus**: the new empty-block test also asserts `verdict.contract.mvp_scope !== null`, which pins the "declared-but-malformed ≠ absent" contract for downstream consumers — exactly what Codex's review and mine both pointed at.

Nits from my prior review (`Object.freeze` on Array vs `Set`, regex-comment for multiline strings) remain non-blocking and were correctly not addressed.

— from claude

PR #180, opened 2026-05-23
feat(hooks): validate Stage A mvp + release_cuts schema in contract graph

KyubumShin:
## What

Stage A implementation phase 1.4a: extend the decomposition contract-graph validator with parser + structural validation for the new `mvp` and `release_cuts[]` top-level fields (landed in PR #178).

- `hooks/lib/mpl-phase-contract-graph.mjs` (+91, -1): `parseMvpObject` + `parseReleaseCuts` helpers; new `validateMvpAndReleaseCuts` invoked from `validatePhaseContractGraph`; exported `ALLOWED_RELEASE_ARTIFACTS` constant.
- `hooks/__tests__/mpl-phase-contract-graph.test.mjs` (+295): 17 new tests under \"Stage A mvp + release_cuts schema\" describe.

Full hook suite: **931/931 pass** (+17, 0 regression).

## Why

PR #178 added the `mvp_scope` field at the goal-contract layer. The decomposer is expected to derive a top-level `mvp` object and optional `release_cuts[]` array on `decomposition.yaml` from that. Without hook-level validation, decomposer drift (missing fields, cross-cut phase overlap, Stage B-only execution modes leaking through) would surface only at runtime — far from the source.

This PR adds the **structural integrity layer**:
- mvp / release_cuts schema parsing
- mvp.phases ⊆ phases; no duplicates
- release_cuts[].phases ⊆ phases; no cross-cut overlap (mvp ∩ cut, cut[i] ∩ cut[j])
- mvp.execution_mode required and Stage-A-only (`sequential`; `contract_skeleton` rejected)
- mvp.artifact / release_cuts[].artifact ∈ `ALLOWED_RELEASE_ARTIFACTS`
- release_cuts[].id unique and never the reserved literal `\"mvp\"`
- release_cuts[].user_approved must be explicit boolean

Backward compatibility: contracts without `mvp` and `release_cuts[]` remain valid (the existing pipeline runs unchanged).

## Design

- Source: `docs/roadmap/stage-a-mvp-cuts-rfc.md` §4.2 (validation requirements), §6 Hook Impact (mpl-phase-contract-graph.mjs row).
- Companion PR (already merged): #178 (`mvp_scope` parser at goal-contract layer).
- Scope deliberately **excludes** two larger items that will land separately in **phase 1.4b**:
  - **Dependency-closure rule (B5)**: no forward or upward `requires`. Needs cross-phase requires analysis across cohort boundaries.
  - **D-Q6 released-cut immutability**: needs `state.release.completed_cut_ids` access at hook runtime (not just single-graph parsing); pairs naturally with state.release subtree (phase 1.6).

## Agent Review Status

This PR is gated on **2 unique agent reviews** (footers `— from <agent>`).

- [ ] from claude
- [ ] from codex

---

### DIFF
--- a/hooks/__tests__/mpl-phase-contract-graph.test.mjs
+++ b/hooks/__tests__/mpl-phase-contract-graph.test.mjs
@@ -165,6 +165,26 @@ describe('Stage A mvp + release_cuts schema', () => {
     assert.equal(graph.mvp.artifact, 'draft_pr');
   });
 
+  it('parses release_cuts items with block-list phases without splitting at nested dashes', () => {
+    // Regression for codex review on 02beb4d: indent-blind `-` detection
+    // misread inner `- phase-2` as a new cut boundary, splitting one valid
+    // cut into two malformed items. Cut item boundaries must respect indent.
+    const text = graphWithMvp('', `release_cuts:
+  - id: cut-a
+    phases:
+      - phase-2
+    user_approved: true
+    artifact: release_manifest`);
+    const graph = parsePhaseContractGraphText(text);
+    assert.equal(graph.release_cuts.length, 1, `cut count: expected 1, got ${graph.release_cuts.length}`);
+    assert.equal(graph.release_cuts[0].id, 'cut-a');
+    assert.deepEqual(graph.release_cuts[0].phases, ['phase-2']);
+    assert.equal(graph.release_cuts[0].user_approved, true);
+    assert.equal(graph.release_cuts[0].artifact, 'release_manifest');
+    const verdict = validatePhaseContractGraph(graph);
+    assert.equal(verdict.valid, true, verdict.issues.join(', '));
+  });
+
   it('parses release_cuts items even when id is not the first key (YAML mapping order is insignificant)', () => {
     // Regression for codex review: previously, parseReleaseCuts only recognized
     // items whose first line was `- id:`, which silently dropped items with
--- a/hooks/__tests__/mpl-phase-contract-graph.test.mjs
+++ b/hooks/__tests__/mpl-phase-contract-graph.test.mjs
@@ -165,6 +165,42 @@ describe('Stage A mvp + release_cuts schema', () => {
     assert.equal(graph.mvp.artifact, 'draft_pr');
   });
 
+  it('parses release_cuts items even when id is not the first key (YAML mapping order is insignificant)', () => {
+    // Regression for codex review: previously, parseReleaseCuts only recognized
+    // items whose first line was `- id:`, which silently dropped items with
+    // any other field order. Validator would then report `valid: true` because
+    // there were no items to check.
+    const text = graphWithMvp('', `release_cuts:
+  - phases: [phase-2]
+    id: cut-a
+    user_approved: true
+    artifact: release_manifest`);
+    const graph = parsePhaseContractGraphText(text);
+    assert.equal(graph.release_cuts.length, 1);
+    assert.equal(graph.release_cuts[0].id, 'cut-a');
+    assert.deepEqual(graph.release_cuts[0].phases, ['phase-2']);
+    assert.equal(graph.release_cuts[0].user_approved, true);
+    assert.equal(graph.release_cuts[0].artifact, 'release_manifest');
+
+    // And the validator MUST still see the same constraints — e.g., overlap
+    // detection must trigger even when ids are not first.
+    const overlapText = graphWithMvp(
+      `mvp:
+  derived_from: goal_contract.mvp_scope
+  phases: [phase-1]
+  execution_mode: sequential
+  artifact: draft_pr`,
+      `release_cuts:
+  - phases: [phase-1]
+    id: cut-a
+    user_approved: true
+    artifact: release_manifest`,
+    );
+    const verdict = validatePhaseContractGraph(parsePhaseContractGraphText(overlapText));
+    assert.equal(verdict.valid, false);
+    assert.ok(verdict.issues.some((i) => i.startsWith('release_cuts:cut-a:phases:overlap:phase-1:already_in:mvp')));
+  });
+
   it('parses release_cuts list with id, phases, user_approved, artifact', () => {
     const text = graphWithMvp('', `release_cuts:
   - id: cut-ext-a
--- a/hooks/__tests__/mpl-phase-contract-graph.test.mjs
+++ b/hooks/__tests__/mpl-phase-contract-graph.test.mjs
@@ -7,6 +7,7 @@ import { fileURLToPath } from 'url';
 import { tmpdir } from 'os';
 
 import {
+  ALLOWED_RELEASE_ARTIFACTS,
   parsePhaseContractGraphText,
   validatePhaseContractGraph,
 } from '../lib/mpl-phase-contract-graph.mjs';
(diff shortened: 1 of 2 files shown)


KyubumShin (commented):
1. **High**: `release_cuts[]` 항목의 `id`가 첫 필드가 아니면 cut 전체가 파싱되지 않고 validator가 통과합니다. `parseReleaseCuts()`는 새 cut을 시작할 때 `- id: ...` 한 형태만 인식합니다(`hooks/lib/mpl-phase-contract-graph.mjs:174`-`178`). 그래서 아래처럼 유효한 YAML 객체 순서만 바뀐 경우 `parsePhaseContractGraphText()`가 `release_cuts: []`로 만들고, `validatePhaseContractGraph()`도 `valid: true`를 반환합니다.

```yaml
release_cuts:
  - phases: [phase-2]
    id: cut-a
    user_approved: true
    artifact: release_manifest
```

이러면 `release_cuts[].phases ⊆ phases`, `user_approved` 필수, artifact enum, cross-cut overlap 같은 이번 PR의 핵심 검증을 전부 우회합니다. YAML mapping field order는 의미가 없고, 기존 goal-contract parser도 list item 내부 `id`가 첫 키가 아니어도 처리하는 회귀 테스트가 이미 있습니다. `parseReleaseCuts()`가 `-` item block을 먼저 나눈 뒤 block 안에서 `id`를 찾거나, 최소한 `release_cuts:`가 present인데 item을 하나도 파싱하지 못한 비-empty block은 invalid로 처리해야 합니다. 회귀 테스트는 `id`를 `phases` 뒤에 둔 형태가 좋습니다.

확인한 테스트:
- `node --test hooks/__tests__/mpl-phase-contract-graph.test.mjs` 통과
- `npm test` 통과 (`931 tests / 0 failures`)

— from codex

KyubumShin (commented):
## Review: feat(hooks) — validate Stage A `mvp` + `release_cuts[]` schema in contract graph

Solid layer-1.4a expansion: 17 new tests, well-structured issue tokens (`mvp:phases:duplicate:<id>`, `release_cuts:cut-a:phases:overlap:phase-1:already_in:mvp`), and a clean `parseMvpObject`/`parseReleaseCuts` + `validateMvpAndReleaseCuts` split. A few items worth surfacing before merge.

### Medium — comment/code contradiction on missing artifact

At `hooks/lib/mpl-phase-contract-graph.mjs:333–337`:

```js
if (cut.artifact === null) {
  // Default applied at decomposer layer; allow absent here but flag explicitly
  // to keep the schema honest — decomposer should emit it.
  issues.push(`release_cuts:${cut.id}:artifact:missing`);
} else if (...) ...
```

The comment says "allow absent here but flag explicitly," but `issues.push` adds it to the verdict's blocking issues — `valid: false` is the result. The two intents are mutually exclusive: either the schema accepts absent artifact (then move the message to a separate `warnings` channel and don't fail the verdict), or it rejects absent (then update the comment to match — "Default applied at decomposer layer **is required** at hook level; absent fails validation").

Recommend: update the comment to match the rejection behavior. The `mvp.artifact:missing` path on line 286 has no such comment and correctly rejects — keep parity.

### Medium — `parseReleaseCuts` assumes `id:` is the first field of each cut item

At `hooks/lib/mpl-phase-contract-graph.mjs:172`:

```js
const idMatch = line.match(/^\s*-\s+id\s*:\s*["']?([^"'\s#]+)["']?/);
```

This treats every `- id: X` line as a new cut boundary. If a cut item is written with fields in a different order (perfectly valid YAML):

```yaml
release_cuts:
  - phases: [phase-2]
    id: cut-a
    user_approved: true
    artifact: release_manifest
```

`parseReleaseCuts` will silently miss the cut entirely (no `- id:` line at the start) — the `- phases:` line doesn't match, so `cur` stays null, and the cut is dropped. The schema test suite always writes `id:` first, so this isn't caught.

Two options:
- (a) Make the parser robust to field order (scan a whole `- ...` item, then extract `id:` from anywhere inside).
- (b) Document the constraint at the parser definition AND add a test that confirms graceful behavior when `id:` isn't first (today: silent drop; preferable: explicit "release_cuts:item:id:missing" error).

This is a parser brittleness rather than a logic bug — the prompt-controlled YAML subset probably always emits `id:` first — but it's worth either pinning or hardening because the failure mode (silent drop) is the worst kind.

### Medium — `mvp.derived_from` is parsed but never validated

`parseMvpObject` reads `derived_from` (line 144) and the test fixture always sets it to `goal_contract.mvp_scope`. But `validateMvpAndReleaseCuts` never inspects it. Per the Stage A RFC §4.2, `derived_from` looks like a provenance marker — is it required to equal `goal_contract.mvp_scope`, or purely informational?

If informational, fine — but the parser shouldn't bother reading it (or add a comment saying "read for future-use only"). If load-bearing, add a validator clause + test, e.g.:

```js
if (mvp.derived_from && mvp.derived_from !== 'goal_contract.mvp_scope') {
  issues.push(`mvp:derived_from:unsupported:${mvp.derived_from}`);
}
```

Quick win: pick one stance, document it in the validator.

### Minor — test coverage gaps

1. **`mvp` block present but `phases:` field entirely missing**: parser returns `phases: []` and validator emits `mvp:phases:missing`. Behavior is correct by inspection, but no test pins it. Add:
   ```js
   it('rejects mvp with no phases field', () => {
     const text = graphWithMvp(`mvp:
     derived_from: goal_contract.mvp_scope
     execution_mode: sequential
     artifact: draft_pr`);
     const verdict = validatePhaseContractGraph(parsePhaseContractGraphText(text));
     assert.equal(verdict.valid, false);
     assert.ok(verdict.issues.includes('mvp:phases:missing'));
   });
   ```

2. **`mvp.phases: []` (explicit empty array)**: same outcome but different YAML shape — worth pinning.

3. **`release_cuts: []` (explicit empty list)**: should be accepted as valid backward-compat — pin it.

4. **Block-list form for `phases`** (`inlineListItems` line 117–125): all tests use inline `phases: [phase-1]`. The block-list code path is untested. Add one test with:
   ```yaml
   phases:
     - phase-1
   ```

### Nit

- `ALLOWED_RELEASE_ARTIFACTS` and `MVP_SCOPE_ARTIFACTS` (from PR #178) hold the identical four values (`draft_pr`, `branch`, `tag`, `release_manifest`) but live in two different files as two different frozen arrays. Future drift between them would silently allow a contract whose artifact is rejected at the goal-contract layer but accepted at the graph layer (or vice versa). Consider exporting one canonical constant from a shared location (or have one re-export the other). Not urgent.

### Verdict

Direction is solid, the schema vocabulary (token format) is well-designed, and the scope is correctly narrowed (excluding B5 dependency closure and D-Q6 immutability — both deferred to 1.4b/1.6 with clear rationale in the PR body). Items 1–3 are real issues with quick fixes; items 4 are test-coverage hygiene that compounds over future iteration. None require redesign.

— from claude

KyubumShin (commented):
## Re-review at `02beb4d` — 1 of 3 medium concerns fixed; 2 still open

Verified the new commit against my prior review.

### Fixed

- **Medium #2 (`parseReleaseCuts` assumed `id:` first)** — fixed cleanly at `hooks/lib/mpl-phase-contract-graph.mjs:151–195`. The new shape splits into `-` item blocks first, then runs `/(?:^|\n)\s*id\s*:\s*["']?([^"'\s#]+)["']?/` over each block — so any field order works. The `-` replacement trick (`${itemStart[1]}  ${itemStart[2]}`) is clever: it preserves the indentation invariant that `nestedScalar`/`inlineListItems` depend on (`^\s+key:`), without rewriting those helpers. New regression test ("parses release_cuts items even when id is not the first key") covers both parse and validator overlap detection with reordered fields. Codex's parallel high-severity finding is also resolved by this change.

### Still open from my prior review

- **Medium #1 — comment/code contradiction on missing artifact** (`mpl-phase-contract-graph.mjs:333–337`). Unchanged. Comment still says "allow absent here but flag explicitly" while `issues.push` blocks the verdict. Pick one — recommend updating the comment to match the rejection (parity with `mvp.artifact:missing` at line 286). One-line fix.

- **Medium #3 — `mvp.derived_from` parsed but never validated**. Unchanged. `parseMvpObject` still reads `derived_from` (line 144) but `validateMvpAndReleaseCuts` never inspects it. Either add a one-line check (`derived_from must equal 'goal_contract.mvp_scope'` or whatever the canonical provenance value is per RFC §4.2) plus one test, or document it as purely informational and add a comment at the parser explaining why it's read but ignored.

- **Minor test coverage gaps** — none added in this push:
  - `mvp` block without `phases:` field
  - explicit `mvp.phases: []`
  - explicit `release_cuts: []` (empty list)
  - block-list form for `phases:` inside an mvp or cut

- **Nit — `ALLOWED_RELEASE_ARTIFACTS` and `MVP_SCOPE_ARTIFACTS` duplication** across `mpl-phase-contract-graph.mjs` and `mpl-goal-contract.mjs`. Unchanged. Still a future-drift risk; safe to defer but worth a follow-up.

### Verdict

The parser-robustness fix is exactly right (better than my suggestion (b) — graceful drop with explicit error — by going straight to suggestion (a) full field-order robustness). The remaining medium items are both one-line changes that don't justify another push round-trip — could be folded into the next commit alongside any other adjustments.

— from claude

KyubumShin (commented):
이전 High였던 `release_cuts[]`의 `id` 필드 순서 문제는 현재 head에서 해결된 것 확인했습니다. 다만 새 splitter에 다른 YAML shape 회귀가 남아 있습니다.

1. **Medium**: `parseReleaseCuts()`가 `phases`의 block-list 항목을 새 `release_cuts[]` item으로 오인합니다. `hooks/lib/mpl-phase-contract-graph.mjs:165`의 `itemStart` 매칭이 들여쓰기 깊이를 보지 않고 모든 `- ...` 라인을 cut item 시작으로 처리해서, 아래처럼 유효한 block-list form이 두 개의 cut으로 쪼개집니다.

```yaml
release_cuts:
  - id: cut-a
    phases:
      - phase-2
    user_approved: true
    artifact: release_manifest
```

재현 결과 `cut-a`는 `phases/user_approved/artifact`가 비어 있는 item이 되고, `- phase-2`는 `id:null`인 별도 item처럼 파싱되어 `release_cuts:cut-a:phases:missing`, `release_cuts:cut-a:user_approved:missing`, `release_cuts:cut-a:artifact:missing`, `release_cuts:id:missing`으로 실패합니다. `inlineListItems()`가 block-list를 지원하도록 들어온 상태라면 이 YAML은 정상 파싱되어야 합니다.

수정 방향은 첫 release cut item의 indent를 기준으로 같은 indent의 `-`만 item boundary로 인정하거나, item block 분리에서도 `phases:` 하위 list indent를 구분하는 방식이 좋겠습니다. 회귀 테스트는 위 block-list `phases` 형태를 그대로 추가하면 잡힙니다.

확인한 테스트:
- `node --test hooks/__tests__/mpl-phase-contract-graph.test.mjs` 통과
- `npm test` 통과 (`932 tests / 0 failures`)

— from codex

KyubumShin (commented):
## Re-review at `7910ff3` — all remaining concerns addressed

Verified each open item from my prior review:

- **Medium #1 (comment/code contradiction on missing artifact)** — fixed at `mpl-phase-contract-graph.mjs:347–352`. New comment matches the rejection behavior with explicit RFC §4.2 reference: "Decomposer MUST emit artifact for every cut … Absence at hook layer means the decomposer skipped the field — that is a contract violation, not a missing default." Parity with `mvp.artifact:missing` (line 286) restored.

- **Medium #3 (`mvp.derived_from` parsed but never validated)** — fixed at `mpl-phase-contract-graph.mjs:17` and `:303–305`. New `ALLOWED_MVP_DERIVED_FROM = ['goal_contract.mvp_scope']` constant + validator clause that treats null as "informational omission" (OK) but any non-null non-canonical value as a drift signal (`mvp:derived_from:unsupported:<value>`). The accept-null-but-reject-drift policy is the right middle ground — it doesn't force every emitter to write the marker, but catches anyone who writes the wrong one. Two new tests pin both branches.

- **Minor test coverage gaps** — all four added:
  - "rejects mvp with no phases field at all"
  - "rejects mvp.phases as an explicit empty inline list"
  - "accepts release_cuts as an explicit empty inline list (backward-compat)" — confirms backward-compat for projects with `release_cuts: []`
  - "parses mvp.phases in block-list form" — exercises the previously-untested code path in `inlineListItems`

- **Nit (`ALLOWED_RELEASE_ARTIFACTS` vs `MVP_SCOPE_ARTIFACTS` duplication)** — fixed at `mpl-phase-contract-graph.mjs:10–14`. Now imported from `mpl-goal-contract.mjs` and re-exported under the graph-layer name with a clear rationale comment ("the set itself is single-sourced … to prevent drift"). Single source of truth.

Bonus correctness fix I didn't flag: `nestedScalar` regex switched from `^\s+` to `^\s*` (line 137) because `topLevelBlock` trims leading whitespace from its output. Without this, the first nested field of a block would silently fail to match. Good catch outside the review scope.

— from claude

KyubumShin (commented):
`7910ff3d` 기준으로 재검토했습니다. Claude가 남겼던 artifact comment, `mvp.derived_from`, artifact constant 중복, 일부 테스트 갭은 이번 head에서 정리된 것 확인했습니다. 다만 이전 Codex 코멘트의 block-list `release_cuts[].phases` 문제는 아직 PR head에 남아 있습니다.

1. **Medium**: `parseReleaseCuts()`가 여전히 nested `phases` block-list 항목을 새 `release_cuts[]` item으로 오인합니다. `hooks/lib/mpl-phase-contract-graph.mjs:168`의 `itemStart`가 모든 들여쓰기의 `- ...`를 item boundary로 처리하기 때문에 아래 유효한 YAML이 실패합니다.

```yaml
release_cuts:
  - id: cut-a
    phases:
      - phase-2
    user_approved: true
    artifact: release_manifest
```

깨끗한 `7910ff3d` worktree에서 재현하면 `release_cuts`가 `cut-a`와 `id:null` 두 item으로 쪼개지고, validator가 `release_cuts:cut-a:phases:missing`, `release_cuts:cut-a:user_approved:missing`, `release_cuts:cut-a:artifact:missing`, `release_cuts:id:missing`을 냅니다. `inlineListItems()`와 새 `mvp.phases` 테스트가 block-list form을 지원하는 방향이라면 cut의 `phases`에도 같은 shape가 통과해야 합니다.

수정은 첫 cut item의 base indent를 잡고 같은 indent의 `-`만 새 cut boundary로 인정하면 충분합니다. 회귀 테스트는 위 `release_cuts[].phases` block-list 케이스를 추가하면 됩니다.

확인한 테스트:
- 깨끗한 `7910ff3d` worktree에서 `node --test hooks/__tests__/mpl-phase-contract-graph.test.mjs` 통과 (`32 tests / 0 failures`)
- 깨끗한 `7910ff3d` worktree에서 `npm test` 통과 (`938 tests / 0 failures`)

— from codex

KyubumShin (commented):
## Re-review at `41e0a18` — block-list `phases:` regression cleanly fixed

Verified the fix at `mpl-phase-contract-graph.mjs:155–212`:

- The old `02beb4d` parser used indent-blind `-` detection over a `topLevelBlock`-trimmed string, so an inner `- phase-2` from a block-list `phases:` field was misread as a new cut boundary — splitting one valid cut into two malformed items.
- The new parser:
  1. Walks raw text directly (avoiding the `topLevelBlock` indent trim that erased the first `-`'s leading whitespace).
  2. Detects `release_cuts:` start line including the inline empty case (`release_cuts: []`).
  3. Reads body lines until the next top-level key (`/^\S/`).
  4. Splits into items only at `-` lines whose indent equals `baseIndent`; deeper `-` lines (block-list children) stay with the current item.
- New test "parses release_cuts items with block-list phases without splitting at nested dashes" exercises the exact failure case and asserts both the parse shape and the validator verdict.

Boundary cases I confirmed by inspection:
- `release_cuts: []` (inline empty) → returns `[]`. ✓
- `release_cuts:` with no body / blank lines / EOF → `bodyLines` empty or no `-` items → returns `[]`. ✓
- `-` at indent shallower than `baseIndent`: doesn't match the `===` predicate, falls into "part of current item" — graceful degradation for malformed YAML; not a regression. ✓
- The previously-tested "id-not-first" regression test still passes because that test uses inline `phases: [...]`, so the inner-dash issue doesn't apply there.

— from claude

PR #181, opened 2026-05-23
feat(phase0): add MVP scope guided-checklist step before Goal Contract freeze

KyubumShin:
## What

Stage A implementation **phase 1.3**: introduce the new **Stage 1.3.5** step in the Phase 0 protocol that elicits the user-declared `mvp_scope` via a guided checklist before the Goal Contract is serialized.

- `commands/mpl-run-phase0.md` (+79 lines): new sub-stage \"MVP Scope Selection (Stage A, optional)\" between Stage 1.3 (user-contract) and Stage 1.4 (Goal Contract freeze). Protocol covers opt-in/opt-out, AC/AX guided checklists, both-empty retry guard, artifact selection. `goal_contract` template gains optional `mvp_scope` field.

Prompt-only change. No code, no tests added/modified. Hook test suite still 938/938 (unchanged by this commit).

## Why

PR #178 added the `mvp_scope` field at the goal-contract parser layer. PR #180 added structural validation of `mvp` / `release_cuts[]` at the decomposition layer. But until this PR, **no protocol step actually elicits `mvp_scope` from the user** — the field is parseable but never populated.

This PR closes that loop. The RFC §10 D-Q1 decision (guided checklist) lands at the only place where the input naturally exists: Phase 0, between user-contract resolution (where AC/AX ids are finalized) and Goal Contract freeze (where they're written).

Sequencing rationale:
- After 1.3 (user-contract): AC/AX ids exist and the user has just confirmed them — fresh in context.
- Before 1.4 (Goal Contract freeze): the freeze step writes `mvp_scope` into the file atomically; no separate write.
- The opt-out branch yields `mvp_scope: null` and the pipeline runs as today.

## Design

- Source: `docs/roadmap/stage-a-mvp-cuts-rfc.md` §10 D-Q1 (guided checklist over id-only or prose-mapped).
- Companion PRs (already merged): #178 (`mvp_scope` parser), #180 (mvp/release_cuts schema validators at decomposition layer).
- Out of scope (separate follow-up):
  - Decomposer-side `mvp.phases` derivation from `mvp_scope` (consumed by mpl-decomposer's prompt).
  - Release-path orchestrator states (`release-gate`, `release-finalize`).
  - State.release subtree.

## Risk

- Pure prompt change. Reviewable as English protocol; no code path can break at runtime from this PR alone.
- Backward compat: users who opt out get exactly today's behavior. mvp_scope absent → goal-contract validator accepts → no decomposer change → no release path.

## Agent Review Status

[...]

PR #179, opened 2026-05-23
feat(decomposer): extend Rule 9 with released-cut immutability (D-Q6)

KyubumShin:
## What

Single one-line extension to `agents/mpl-decomposer.md` Rule 9 (RECOMPOSE-MODE / APPEND-MODE). Adds the **Stage A released-cut immutability constraint** as a forward-reference to the Stage A RFC.

- `agents/mpl-decomposer.md`: Rule 9 now states that when recompose runs against a graph carrying Stage A release fields, `mvp.phases` and `release_cuts[id].phases` (for `id ∈ state.release.completed_cut_ids`) are immutable. New phases go to a new cut or non-cut tail.

## Why

Resolves D-Q6 from the Stage A RFC at the agent-prompt layer. A released cut's `release-manifest.json` has already been shipped externally (PR / branch / tag); mutating its phase membership would invalidate the shipped artifact. The constraint must reach the decomposer agent before any release-path hook lands so that the agent never *attempts* the mutation.

This PR is intentionally tiny (one-paragraph extension) and **independent of PR #178** (the `mvp_scope` parser). The agent prompt cites the RFC, which lives in main since #177.

## Design

- Source: `docs/roadmap/stage-a-mvp-cuts-rfc.md` §10 D-Q6 (\"released cut phase membership is immutable\") and §6 Hook Impact (decomposer.md row).
- Enforcement complement: the corresponding *hook-level* block (validator rejecting decomposition writes that mutate a released cut's phase set) is intentionally deferred to **Phase 1.4** (`hooks/lib/mpl-phase-contract-graph.mjs` extension). That depends on the `mvp` / `release_cuts[]` schema parser landing first via PR #178.
- Forward-reference rationale: the RFC is the SSOT and is already in main; the agent prompt can cite it now to lock in the constraint before code enforces it. This is the same pattern used by other Stage A items (the prompt knows about `release-gate` / `release-finalize` even though the orchestrator states do not exist yet).

## Agent Review Status

This PR is gated on **2 unique agent reviews** (footers `— from <agent>`).

- [ ] from claude
- [ ] from codex

---
_Awaiting reviews. Merge once the gate is satisfied._

PR #177, opened 2026-05-23
docs: Stage A MVP/release-cuts analysis + RFC

KyubumShin:
## What

- New `docs/roadmap/parallelism-mvp-analysis.md` (390 lines) — diagnostic + design analysis
- New `docs/roadmap/stage-a-mvp-cuts-rfc.md` (379 lines) — implementation-ready Stage A RFC

## Why

체감 두 문제 — (1) 낮은 병렬성, (2) MVP까지의 긴 시간 — 을 분석한 결과 양쪽 다 사실로 확인됨. 기존 pipeline의 primitives (`interface_contract`, `execution_tiers`, 병렬 executor, TODO-level concurrency)는 충분한데 default가 보수적이고 모든 hard gates가 후방에 몰려 있어 partial artifact가 user-visible해진 적이 한 번도 없음.

Stage A RFC는 user-declared `mvp` + `release_cuts[]` + 새 release path (`release-gate(cut_id)` → `release-finalize(cut_id)`)로 이 문제를 해소하면서, 기존 `phase3-gate` / `phase5-finalize`의 whole-pipeline gate / whole-goal closure semantics는 *약화 없이* 유지하도록 설계됨. MVP 실행은 Stage A에서 sequential로만; 병렬 MVP (`contract_skeleton`)는 Stage B로 미룸 — critical path 안전이 우선.

## Design

이 PR이 곧 설계 문서임 (self-referential):

- **`docs/roadmap/parallelism-mvp-analysis.md`** — 진단 + 4개 제안 평가 + 두 차례 review rebaseline (MVP/cut 분리, 용어 일관성, release path vs whole-pipeline finalize 분리) + Skeleton verification & Reconcile failure policy 결정 (Stage B 사전 결정) + Safer Conflict Rule + Stage A/B/C 권장 시퀀스 + Open Questions (다수 resolved) + source references.
- **`docs/roadmap/stage-a-mvp-cuts-rfc.md`** — Stage A 구현 spec:
  - Schema: `goal_contract.mvp_scope`, `decomposition.mvp`, `release_cuts[]`, `state.release` subtree, release artifacts under `.mpl/mpl/releases/{cut_id}/`
  - Orchestrator: 2-state release path, cohort-aware `phase2-sprint` completion (line 415), scoped Hard 1/2/3 with surfaced Hard 2 fallback, `phase3-gate`/`phase5-finalize` unchanged
  - Hook impact: detailed file-by-file changes including new `mpl-require-export-manifest.mjs`
[...]

PR #182, opened 2026-05-23
feat(decomposer): derive mvp cohort from goal_contract.mvp_scope (Stage A)

KyubumShin:
## What

Stage A implementation **phase 1.2**: extend the decomposer agent to emit the Stage A `mvp` object on `decomposition.yaml` by mechanically mapping `goal_contract.mvp_scope` AC/AX ids to phase ids via `goal_trace`.

- `agents/mpl-decomposer.md` (+84 lines, prompt-only):
  - New **Step 12: MVP cut derivation** in Reasoning_Steps
  - Output_Schema gains `mvp` (single object, optional) + `release_cuts: []` (Stage A always empty) top-level fields

## Why

This closes the end-to-end Stage A loop:

```
PR #178: goal-contract.mvp_scope parser
PR #181: interview elicits mvp_scope from user
THIS  → decomposer derives decomposition.mvp from mvp_scope
PR #180: validators reject malformed mvp / release_cuts
Phase 1.6 (next): orchestrator routes MVP cohort via release-gate/release-finalize
```

Before this PR, `mvp_scope` was a dead-end field — parsed and validated but never consumed. After this PR, the decomposer materializes `decomposition.mvp.phases` so the Phase 1.6 orchestrator has something concrete to route on.

Per RFC §3.4 / §10 D-Q4, derivation is **mechanical id-set mapping**, not semantic inference. The decomposer never decides what belongs in MVP — it only walks the user's declaration onto the existing phase graph.

## Design

- Source: `docs/roadmap/stage-a-mvp-cuts-rfc.md` §4.2 (mvp schema), §3.4 (user-declared not inferred), §10 D-Q4 (artifact stance), §10 D-Q2 (cuts auto-proposal is Stage B).
- Coverage check: every `mvp_scope` id MUST be covered by at least one phase's `goal_trace`. Gap → HIGH risk + `go_no_go: NOT_READY`. User must re-interview or recompose.
- Hard-coded narrowing (intentional):
  - `mvp.execution_mode = \"sequential\"` (Stage A only; Stage B `contract_skeleton` deferred)
  - `release_cuts = []` (Stage A; decomposer-auto-proposal is Stage B per RFC §10 D-Q2)
- Companion PRs (all merged): #177 (RFC), #178 (`mvp_scope` parser), #180 (contract-graph validators), #181 (interview elicitation), #179 (Rule 9 immutability for recompose).

## Risk

- Pure prompt change. Hook test suite still 939/939 (unchanged by this commit).
- Backward compat: when `goal_contract.mvp_scope` is absent, decomposer omits both `mvp` and `release_cuts` and the existing pipeline runs as today.
- Coverage gap (HIGH risk + NOT_READY) is the user-protection guard: prevents an MVP cut that secretly doesn't deliver the declared AC/AX from being scheduled.

## Out of scope (separate phase)

- Release-path orchestrator states (`release-gate`, `release-finalize`) — Phase 1.6
- `state.release` subtree + cohort-aware `phase2-sprint` completion — Phase 1.6
- Dependency-closure rule + D-Q6 immutability hook enforcement — Phase 1.4b
- Auto-proposed extension cuts — RFC §10 D-Q2 Stage B

## Agent Review Status

- [ ] from claude
- [ ] from codex

---

### DIFF
--- a/agents/mpl-decomposer.md
+++ b/agents/mpl-decomposer.md
@@ -300,6 +300,62 @@ disallowedTools: Bash,Task,WebFetch,WebSearch,NotebookEdit
     Step 11: Execution tiers
       - Group phases by dependency level (topological tiers).
       - pp_core phases: parallel = false (sequential). Others: parallel = true if no file overlap.
+
+    Step 12: MVP cut derivation (Stage A, only when goal_contract.mvp_scope is present)
+      When `goal_contract.mvp_scope` is ABSENT, **skip this step entirely**:
+      omit both `mvp` and `release_cuts` from the output. The pipeline runs
+      as today with no Stage A release path.
+
+      When `goal_contract.mvp_scope` IS present, derive `mvp.phases` as the
+      set of phase ids whose `goal_trace` intersects the user-declared
+      MVP id set. This is **mechanical id-set mapping** (NOT semantic
+      inference). Per RFC §3.4 / §10 D-Q4, the decomposer MUST NOT infer
+      MVP scope from anything other than `goal_contract.mvp_scope`.
+
+      ```
+      mvp_ac_ids = goal_contract.mvp_scope.acceptance_criteria  # set of AC ids
+      mvp_ax_ids = goal_contract.mvp_scope.variation_axes        # set of AX ids
+      mvp_target_ids = mvp_ac_ids ∪ mvp_ax_ids
+
+      mvp_phases = []
+      for phase in phases (in execution_tiers order):
+        phase_ids = phase.goal_trace.acceptance_criteria ∪ phase.goal_trace.variation_axes
+        if phase_ids ∩ mvp_target_ids != ∅:
+          mvp_phases.append(phase.id)
+
+      # Coverage check: every mvp_scope id must be covered by at least one phase
+      # in mvp_phases. If not, the user's MVP cannot be realized as decomposed.
+      covered_ac = ⋃ phase.goal_trace.acceptance_criteria for phase in mvp_phases
+      covered_ax = ⋃ phase.goal_trace.variation_axes        for phase in mvp_phases
+      missing_ac = mvp_ac_ids - covered_ac
+      missing_ax = mvp_ax_ids - covered_ax
+      if missing_ac or missing_ax:
+        risk_assessment.risks.append({
+          id: "STAGE_A_MVP_COVERAGE_GAP",
+          severity: "HIGH",
+          title: "MVP scope ids not covered by any phase's goal_trace",
+          description: `Missing AC: ${missing_ac.join(',') or 'none'}; Missing AX: ${missing_ax.join(',') or 'none'}. Either revise goal_contract.mvp_scope to drop these ids, add phases that cover them, or extend an existing phase's goal_trace.`,
+          mitigation: "Re-interview to revise mvp_scope, or recompose with additional phases.",
+        })
+        risk_assessment.go_no_go = "NOT_READY"  # block until coverage resolves
+
+      # Cross-cut overlap check (Stage A: decomposer emits ZERO release_cuts,
+      # so no overlap possible from this step. Stage B may auto-propose cuts;
+      # at that point the contract-graph validator (PR #180) catches overlap.)
+
+      Emit on output:
+        mvp = {
+          derived_from: "goal_contract.mvp_scope",
+          phases: mvp_phases,
+          execution_mode: "sequential",      # Stage A only allows sequential
+          artifact: goal_contract.mvp_scope.artifact,
+        }
+        release_cuts = []
+        # Stage A: decomposer does NOT auto-propose extension cuts.
+        # The orchestrator runs MVP cohort first via release-gate/release-finalize
+        # (Phase 1.6, separate landing), then extension phases via existing
+        # execution_tiers. Auto-proposal of cuts is RFC §10 D-Q2 Stage B work.
+      ```
   </Reasoning_Steps>
 
   <Output_Schema>
@@ -501,6 +557,34 @@ disallowedTools: Bash,Task,WebFetch,WebSearch,NotebookEdit
       - tier: number
         phases: [string]
         parallel: boolean
+
+    # Stage A: MVP cohort + release cuts. OPTIONAL — emit only when
+    # `goal_contract.mvp_scope` is present (see Step 12 derivation rules).
+    # When absent, omit both `mvp` and `release_cuts` entirely; the pipeline
+    # runs as today with no Stage A release path.
+    #
+    # The Stage A validators (hooks/lib/mpl-phase-contract-graph.mjs,
+    # landed in PR #180) check:
+    #   - mvp.phases ⊆ phases[], no duplicates
+    #   - mvp.execution_mode == "sequential" (Stage A only)
+    #   - mvp.artifact ∈ {draft_pr, branch, tag, release_manifest}
+    #   - release_cuts[].id unique, not "mvp"; phases ⊆ phases[]
+    #   - no cross-cut phase overlap (mvp ∩ cut, cut[i] ∩ cut[j])
+    mvp:
+      derived_from: "goal_contract.mvp_scope"
+      phases: [string]               # ids whose goal_trace covers mvp_scope AC/AX
+      execution_mode: "sequential"    # Stage A: always sequential
+      artifact: "draft_pr | branch | tag | release_manifest"  # from mvp_scope.artifact
+
+    release_cuts: []
+      # Stage A: decomposer emits EMPTY release_cuts[] when mvp is present.
+      # Auto-proposal of extension cuts is RFC §10 D-Q2 Stage B work.
+      # When present, each entry has shape:
+      #   - id: string                  # unique within release_cuts; never "mvp"
+      #     phases: [string]            # disjoint from mvp.phases and other cuts
+      #     proposed_by: "mpl-decomposer"
+      #     user_approved: boolean      # planning-stage HITL writes; runtime never mutates
+      #     artifact: "draft_pr | branch | tag | release_manifest"
 
     decomposition_rationale: string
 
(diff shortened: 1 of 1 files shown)


KyubumShin (commented):
검토했습니다. 추가로 발견된 문제 없음. LGTM.

확인한 내용:
- `goal_contract.mvp_scope`가 없을 때 `mvp`/`release_cuts`를 모두 생략하는 경로가 기존 backward-compat 계약과 맞습니다.
- `mvp_scope`가 있을 때 `goal_trace`와의 기계적 id-set intersection으로 `mvp.phases`를 만드는 방식이 RFC의 “user-declared, not inferred” 방향과 맞습니다.
- Stage A에서 `release_cuts: []`만 emit하고 auto-proposed cuts를 Stage B로 미루는 범위도 PR 설명/RFC와 일관됩니다.

테스트: `npm test` 통과 (`939 tests / 0 failures`).

— from codex

KyubumShin (commented):
## Review: feat(decomposer) — derive mvp cohort from goal_contract.mvp_scope (Stage A)

Prompt-only change that finally closes the Stage A loop — mvp_scope now has a downstream consumer. The mechanical id-set mapping (RFC §3.4 / §10 D-Q4) is correctly implemented as "phase intersects MVP id set → pull phase fully into MVP" (matching RFC §12's binary-membership risk mitigation). A few items worth surfacing before merge.

### Medium — verify `phase.goal_trace.variation_axes` exists in the existing schema

Step 12's mapping reads from both fields:

```
phase_ids = phase.goal_trace.acceptance_criteria ∪ phase.goal_trace.variation_axes
```

The MVP scope can include variation_axes ids (per `mvp_scope.variation_axes` in PR #178). But does the existing `goal_trace` schema on each phase include `variation_axes`? Looking at the decomposer's prior Output_Schema (lines above the diff), the typical `goal_trace` shape is `{ acceptance_criteria: [...], covers: [...] }` — I don't see `variation_axes` documented anywhere as a per-phase field.

If `phase.goal_trace.variation_axes` doesn't exist on existing phases, then:
- `mvp_target_ids` containing any AX id will never match any phase
- The coverage check will always emit `missing_ax: [<every-ax-id-in-mvp_scope>]`
- Every contract with an AX in `mvp_scope` will be NOT_READY

This is a real silent-no-op failure. Either:
- (a) The schema already has `variation_axes` in `goal_trace` — please point me at it in the existing prompt
- (b) The schema doesn't — this PR needs to also extend the per-phase `goal_trace` to include `variation_axes`, and the prior decomposition step (Step 10? whichever fills goal_trace) needs to populate it

Without (b) landed, the entire AX side of MVP derivation is dead code. Worth a quick verification before merge.

### Medium — dependency-closure check is missing (and the PR body acknowledges it, but the consequence is worth flagging)

The PR body correctly defers B5 dependency-closure to Phase 1.4b. But the consequence of *not* having it in this PR is:

An MVP phase P_mvp can have `requires.from_phase: P_dep` where P_dep is NOT in mvp_phases. When Phase 1.6 lands and the orchestrator routes the MVP cohort through `release-gate(mvp) → release-finalize(mvp)`, the cohort runs *before* the extension phases. So P_mvp would build against P_dep's contract — but P_dep hasn't executed yet. The MVP gate either fails at Hard 1 (build) or fails at Hard 3 (contract verification, since `interface_contract.produces` from P_dep was never produced).

So this PR, together with the future Phase 1.6 orchestrator, can in principle ship a Stage A decomposition that *will fail at runtime*. The user-protection guard for *coverage* (NOT_READY on missing AC/AX) exists; the equivalent guard for *dependency closure* doesn't. Recommend either:
- (a) Adding a minimal dependency-closure check to Step 12 in this PR (one set-difference operation) so MVP can never ship with unsatisfied requires
- (b) Adding a HIGH risk + NOT_READY clause to Step 12 that explicitly says "this is dependent on Phase 1.4b; until then, MVP cohort may have unsatisfied requires"

Option (a) is cheap and would let Phase 1.6 land without the lurking failure mode. Suggested logic:

```
mvp_requires = ⋃ phase.requires.from_phase for phase in mvp_phases
missing_deps = mvp_requires - {phase.id for phase in mvp_phases}
if missing_deps:
  risk_assessment.risks.append({
    id: "STAGE_A_MVP_DEPENDENCY_GAP",
    severity: "HIGH",
    description: f"MVP phases require non-MVP phases: {missing_deps}. ...",
  })
  risk_assessment.go_no_go = "NOT_READY"
```

The B5 hook-level enforcement (1.4b) is the durable solution, but a transient decomposer-level check would close the door for the next several phases of landing.

### Minor — YAML formatting nit in Output_Schema

The Output_Schema block has:

```yaml
release_cuts: []
  # Stage A: decomposer emits EMPTY release_cuts[] when mvp is present.
  ...
  #   - id: string                  # unique within release_cuts; never "mvp"
  #     phases: [string]
```

The `release_cuts: []` on one line followed by an indented comment block is a YAML-formatting hazard: a strict YAML reader would treat `[]` as the value (correct) but a casual reader sees the indent and assumes the items are nested below. Two fixes either of which works:
- (a) Drop the explicit `[]` and let the comment block describe the field's allowed shape: `release_cuts:  # Stage A: MUST be omitted or empty array. Stage B entry shape: ...`
- (b) Unindent the comment block to the same level as `release_cuts:` so it reads as a sibling comment, not a child.

Schema docs above similar fields (e.g., `execution_tiers`) follow one of these patterns; pick whichever is closer.

### Minor — empty `mvp_phases` emits both a risk AND triggers validator rejection

Edge case: when `goal_contract.mvp_scope` is non-empty but no phase's goal_trace intersects it (e.g., decomposer hasn't covered any of the MVP ids yet), `mvp_phases` is empty. Then:
- Step 12 emits `mvp.phases = []` AND adds `STAGE_A_MVP_COVERAGE_GAP` risk + NOT_READY
- PR #180's validator independently rejects `mvp:phases:missing`

Both block correctly, but the diagnostics are redundant. Not a bug — just worth a one-line note in Step 12 confirming the intent: "Note: when mvp_phases is empty, the validator's mvp:phases:missing will also fire. The Step 12 NOT_READY is the user-facing diagnostic; the validator is the safety net."

### Nit — `release_cuts: []` always-emit vs omit-when-mvp-absent

Step 12 says:
> When `goal_contract.mvp_scope` is ABSENT, **skip this step entirely**: omit both `mvp` and `release_cuts` from the output.

But the Output_Schema comment above says `release_cuts: []` is always emitted (when mvp present, it's empty). These are consistent (both: emit only when mvp_scope present). PR #180's validator accepts both `release_cuts: []` and absent `release_cuts`. So the choice between always-emit-empty and omit is purely stylistic — current PR emits when mvp_scope present, which is fine. No fix needed; just confirming the rule is intentional.

### Verdict

The mapping logic is right and the conditional-skip backward-compat path is correctly stated. The biggest pre-merge question is whether `phase.goal_trace.variation_axes` exists in the schema — if not, this PR has a silent gap on the AX path. The dependency-closure question is acknowledged but worth re-examining whether the transient decomposer-level check is cheap enough to add here vs. waiting for 1.4b. Minor formatting items are quick fixes.

— from claude

KyubumShin (commented):
## Re-review at `40f5751` — all prior concerns addressed

Verified each item from my earlier review:

- **Medium #1 (verify `phase.goal_trace.variation_axes` exists)** — fixed with an explicit schema-reference block right above the mapping pseudocode:
  > Schema reference: each phase already carries `goal_trace.acceptance_criteria[]` and `goal_trace.variation_axes[]` under the per-phase Output_Schema … Both fields are REQUIRED to be set by every phase (per the existing goal_trace contract, enforced by `mpl-require-goal-trace.mjs`), so both sides of the intersection below are well-defined.

  Confirms the AX side is not silent dead code, with a pointer to the hook that enforces the upstream contract.

- **Medium #2 (dependency-closure gap)** — fixed by adding "Guard 2: DEPENDENCY CLOSURE" exactly along the lines suggested:
  ```
  mvp_id_set = {phase.id for phase in mvp_phases}
  mvp_requires = ⋃ phase.interface_contract.requires.from_phase
                   for phase in mvp_phases (skip entries without from_phase — they are baseline refs)
  missing_deps = mvp_requires - mvp_id_set
  if missing_deps:
    risk_assessment.risks.append({ id: "STAGE_A_MVP_DEPENDENCY_GAP", severity: "HIGH", ... })
    risk_assessment.go_no_go = "NOT_READY"
  ```

  Plus a clear explanation of the transient nature (until Phase 1.4b's B5 hook), the runtime consequence ("release-gate(mvp) would have unsatisfied requires"), and the explicit "skip entries without from_phase — they are baseline refs" detail that prevents false positives against baseline refs. This is the right shape — surfaces the gap to the user without blocking the upstream RFC sequencing.

- **Minor #3 (YAML formatting on `release_cuts: []` + indented comments)** — fixed by inverting the order: comment block now precedes the declaration, so the indented comments naturally sit above `release_cuts: []` rather than under it. Reads unambiguously now.

- **Minor #4 (empty `mvp_phases` redundancy)** — fixed with the inline note inside Guard 1:
  > (When mvp_phases ends up empty, the validator's `mvp:phases:missing` in PR #180 will ALSO reject; that is the safety net. This Step 12 NOT_READY is the user-facing diagnostic with the missing-id list.)

  Documents the intent clearly: Step 12 is the user-facing diagnostic, validator is the safety net.

The Guard 1 / Guard 2 framing is a nice readability bonus — makes it obvious that there are two distinct checks and what each defends against.

— from claude

KyubumShin (commented):
`40f5751` 기준으로 재검토했습니다. 추가로 발견된 문제 없음. LGTM.

확인한 내용:
- `goal_trace.variation_axes[]`가 per-phase Output_Schema와 `mpl-require-goal-trace.mjs` parser/validator에 실제로 존재하므로 AX 기반 MVP 매핑 경로가 살아 있습니다.
- Claude가 지적한 transient dependency-closure gap은 Step 12의 `STAGE_A_MVP_DEPENDENCY_GAP` HIGH/NOT_READY guard로 반영됐습니다.
- `release_cuts: []` formatting도 주석 위치가 정리되어 YAML shape 혼동이 줄었습니다.
- empty `mvp_phases`일 때 Step 12 diagnostic과 PR #180 validator safety net이 중복될 수 있다는 점도 주석으로 명시됐습니다.

테스트: `npm test` 통과 (`939 tests / 0 failures`).

— from codex

PR #183, opened 2026-05-23
feat(hooks): enforce dependency-closure (B5) and released-cut immutability (D-Q6)

KyubumShin:
## What

Stage A implementation **phase 1.4b**: extend the contract-graph validator and require hook with the two checks deferred from PR #180 (Phase 1.4a):

- **B5 dependency-closure** (lib validator): no forward or upward `requires` between Stage A cohorts
- **D-Q6 released-cut immutability** (require hook): once `state.release.completed_cut_ids` lists a cut, its phase list is frozen

### Files changed

- `hooks/lib/mpl-phase-contract-graph.mjs` (+58, -3): new `validateDependencyClosure` invoked from `validateMvpAndReleaseCuts`
- `hooks/mpl-require-phase-contract-graph.mjs` (+82, +0): new `validateReleasedCutImmutability` invoked alongside graph validator. Reads `.mpl/state.json` for `state.release.completed_cut_ids`, diffs prior `.mpl/mpl/decomposition.yaml` against new write
- `hooks/__tests__/mpl-phase-contract-graph.test.mjs` (+233, +0): 12 new tests (7 dep-closure unit, 5 D-Q6 hook integration). 1 prior fixture (block-list parser regression) updated to include mvp so its cut's dependency resolves through the new closure check

Full hook suite: **950/950 pass** (+11 net, 0 regression).

## Why

PR #180 (Phase 1.4a) deferred two checks intentionally — they needed cross-phase requires analysis and `state.release` runtime access respectively. Without them:

- An MVP cohort could ship with `requires.from_phase` pointing at a non-MVP phase. The MVP gate would fail at runtime (Phase 1.6 release-gate) far from the source. The decomposer's transient guard from Phase 1.2 Step 12 surfaces it but only at decomposition time; this PR is the durable check at write time.
- A recompose could silently mutate `mvp.phases` after `release-finalize(mvp)` has shipped — invalidating the released artifact. PR #179 added the agent-prompt rule (Rule 9 extension); this PR is the hook-level enforcement.

## Design

- **Source**: `docs/roadmap/stage-a-mvp-cuts-rfc.md` §4.2 (B5 dependency-closure), §10 D-Q6 (released-cut immutability), §5.4.2 (cut activation/ordering — Stage A serializes cuts per declared array order)
- **Companion PRs (merged)**: #178 (`mvp_scope` parser), #179 (Rule 9 immutability prompt), #180 (Phase 1.4a schema validators), #181 (interview elicitation), #182 (decomposer mvp derivation + transient deps guard)
- **Forward-compat**: D-Q6 hook is a no-op until `state.release` exists (Phase 1.6+). The check is wired now so 1.6 doesn't need a second visit to this layer.

## Issue tokens

| Token | Severity | Cause |
|---|---|---|
| `mvp:phases:<id>:requires:outside_cohort:<dep>` | block | MVP phase requires non-MVP phase |
| `release_cuts:<cutId>:phases:<id>:requires:outside_cohort:<dep>` | block | cut phase requires later cut or non-cohort phase |
| `released_cut:<id>:phases:mutated:old=[...]:new=[...]` | block | recompose changed a released cut's phase list |
| `released_cut:<id>:removed_from_graph` | block | recompose removed a released cut entirely |

## Risk

- Pure validator + hook extension. Backward compatibility: pre-Stage-A graphs (no mvp/release_cuts) skip dep-closure; pre-Phase-1.6 projects (no `state.release` in state.json) skip immutability.
- Block-list parser regression fixture (PR #180 codex review) updated to include mvp — its prior shape happened to violate the new dep-closure rule. Behaviour change is intentional and explicit (test comment cites Phase 1.4b).

## Out of scope (separate phase)

- Orchestrator release-path states (`release-gate`, `release-finalize`) — Phase 1.6
- `state.release` subtree write path — Phase 1.6
- Cohort-aware `phase2-sprint` completion — Phase 1.6

## Agent Review Status

- [ ] from claude
- [ ] from codex

---

### DIFF
--- a/hooks/__tests__/mpl-phase-contract-graph.test.mjs
+++ b/hooks/__tests__/mpl-phase-contract-graph.test.mjs
@@ -588,6 +588,100 @@ execution_tiers:`);
     const verdict = validatePhaseContractGraph(parsePhaseContractGraphText(validGraph()));
     assert.equal(verdict.valid, true, verdict.issues.join(', '));
   });
+
+  it('rejects approved cut whose require resolves only into a DECLINED earlier cut (RFC B8)', () => {
+    // Codex high-severity: a declined cut (user_approved: false) is NOT a
+    // release-path cohort — its phases run as non-cut tail, AFTER all
+    // release-path cuts. So a later approved cut MUST NOT count a declined
+    // cut's phases as "already materialized."
+    //
+    // Fixture: cut-a declined with [phase-1]; cut-b approved with [phase-2].
+    // phase-2 requires phase-1. At cut-b's release-gate, phase-1 doesn't
+    // exist yet (cut-a is tail, runs later). Must be flagged.
+    const text = dcGraph('', `release_cuts:
+  - id: cut-a
+    phases: [phase-1]
+    user_approved: false
+    artifact: release_manifest
+  - id: cut-b
+    phases: [phase-2]
+    user_approved: true
+    artifact: release_manifest`);
+    const verdict = validatePhaseContractGraph(parsePhaseContractGraphText(text));
+    assert.equal(verdict.valid, false);
+    assert.ok(
+      verdict.issues.includes('release_cuts:cut-b:phases:phase-2:requires:outside_cohort:phase-1'),
+      `expected declined-cut-as-tail to surface dep-closure issue; got: ${verdict.issues.join(', ')}`,
+    );
+  });
+
+  it('skips dep-closure for declined cuts themselves (tail semantics, final phase3-gate covers)', () => {
+    // A declined cut's OWN phase requires are not release-path concerns —
+    // they run as tail and are verified at final phase3-gate. The dep-
+    // closure check must not produce issues for declined cuts themselves.
+    //
+    // Fixture: mvp = [phase-1]; cut-a declined with [phase-2]. phase-2
+    // requires phase-1 — which is in mvp. But even if cut-a's phase-2 had
+    // required some non-cohort phase, dep-closure would skip the check
+    // because cut-a is declined.
+    const text = dcGraph(
+      `mvp:
+  derived_from: goal_contract.mvp_scope
+  phases: [phase-1]
+  execution_mode: sequential
+  artifact: draft_pr`,
+      `release_cuts:
+  - id: cut-a
+    phases: [phase-2]
+    user_approved: false
+    artifact: release_manifest`,
+    );
+    const verdict = validatePhaseContractGraph(parsePhaseContractGraphText(text));
+    assert.equal(verdict.valid, true, verdict.issues.join(', '));
+  });
+
+  it('parser: requires_from_phases extracts only entries with from_phase, ignoring baseline-only requires', () => {
+    // Pin the parser contract that validateDependencyClosure depends on:
+    // a `requires:` list mixing a from_phase entry and a baseline-only entry
+    // (no from_phase key) must yield ONLY the from_phase id.
+    //
+    // Fixture: phase-2 requires both phase-1 (from_phase) AND a baseline
+    // artifact (no from_phase). Only phase-1 is subject to closure check.
+    // Therefore mvp = [phase-2] alone still flags phase-1 as out-of-cohort
+    // (the from_phase entry), but does NOT flag the baseline entry.
+    const text = validGraph()
+      .replace(
+        `interface_contract:
+      requires:
+        - type: artifact
+          name: bootstrap
+          from_phase: phase-1
+      produces: []`,
+        `interface_contract:
+      requires:
+        - type: artifact
+          name: bootstrap
+          from_phase: phase-1
+        - type: artifact
+          name: baseline_only_dep
+      produces: []`,
+      )
+      .replace('execution_tiers:', `mvp:
+  derived_from: goal_contract.mvp_scope
+  phases: [phase-2]
+  execution_mode: sequential
+  artifact: draft_pr
+execution_tiers:`);
+    const verdict = validatePhaseContractGraph(parsePhaseContractGraphText(text));
+    // Should fail ONLY because of phase-1 (the from_phase entry) being
+    // out of cohort. The baseline_only_dep entry must NOT appear.
+    assert.equal(verdict.valid, false);
+    assert.ok(verdict.issues.includes('mvp:phases:phase-2:requires:outside_cohort:phase-1'));
+    assert.ok(
+      !verdict.issues.some((i) => i.includes('baseline_only_dep')),
+      `baseline-only require must not be subject to closure; got: ${verdict.issues.join(', ')}`,
+    );
+  });
 });
 
 describe('mpl-require-phase-contract-graph hook integration', () => {
@@ -707,6 +801,27 @@ describe('mpl-require-phase-contract-graph hook integration', () => {
   artifact: draft_pr`);
     const r = runHook(mutated);
     assert.equal(r.continue, true);
+  });
+
+  it('D-Q6: blocks reorder-only mutations of a released cut (Rule 9 alignment)', () => {
+    // PR #179's Rule 9 forbids "add a phase id to, remove one from, OR
+    // reorder phases within a released cut's .phases list." `phasesEqual`
+    // is order-sensitive by design — pin that contract.
+    const priorMvp = `mvp:
+  derived_from: goal_contract.mvp_scope
+  phases: [phase-1, phase-2]
+  execution_mode: sequential
+  artifact: draft_pr`;
+    writeFileSync(join(tmp, '.mpl', 'mpl', 'decomposition.yaml'), graphWithMvp(priorMvp));
+    seedStateWithReleased(['mvp']);
+    const reordered = graphWithMvp(`mvp:
+  derived_from: goal_contract.mvp_scope
+  phases: [phase-2, phase-1]
+  execution_mode: sequential
+  artifact: draft_pr`);
+    const r = runHook(reordered);
+    assert.equal(r.decision, 'block');
+    assert.match(r.reason, /released_cut:mvp:phases:mutated/);
   });
 
   it('D-Q6: no-op for projects whose state.json lacks state.release (pre-Phase-1.6)', () => {
--- a/hooks/__tests__/mpl-phase-contract-graph.test.mjs
+++ b/hooks/__tests__/mpl-phase-contract-graph.test.mjs
@@ -169,12 +169,23 @@ describe('Stage A mvp + release_cuts schema', () => {
     // Regression for codex review on 02beb4d: indent-blind `-` detection
     // misread inner `- phase-2` as a new cut boundary, splitting one valid
     // cut into two malformed items. Cut item boundaries must respect indent.
-    const text = graphWithMvp('', `release_cuts:
+    //
+    // The fixture also includes mvp=[phase-1] so cut-a's phase-2 dependency
+    // on phase-1 resolves through the earlier cohort (mvp) for the new
+    // dependency-closure validator (Phase 1.4b).
+    const text = graphWithMvp(
+      `mvp:
+  derived_from: goal_contract.mvp_scope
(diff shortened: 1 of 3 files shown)


KyubumShin (commented):
1. **High**: dependency-closure가 `user_approved:false`인 앞선 cut을 이미 실행된 cohort처럼 취급합니다. `validateDependencyClosure()`는 cut을 순서대로 돌면서 검증이 끝난 모든 cut의 phase를 `earlierPhases`에 추가합니다(`hooks/lib/mpl-phase-contract-graph.mjs:423`-`436`). 그런데 RFC B8은 `user_approved:false` cut의 phase가 release path를 타지 않고 non-cut tail로 실행된다고 정의합니다(`docs/roadmap/stage-a-mvp-cuts-rfc.md:99`-`103`). 따라서 declined cut은 뒤의 approved cut의 dependency를 만족시키면 안 됩니다.

재현 예:

```yaml
release_cuts:
  - id: cut-a
    phases: [phase-1]
    user_approved: false
    artifact: release_manifest
  - id: cut-b
    phases: [phase-2]
    user_approved: true
    artifact: release_manifest
# phase-2 requires.from_phase: phase-1
```

현재 validator는 이 graph를 `valid: true`로 통과시킵니다. 하지만 runtime에서는 `cut-a`가 release-gate/release-finalize를 거치지 않고 tail로 밀리므로, `cut-b`의 release-gate 시점에는 `phase-1` 산출물이 아직 없습니다. 기대 결과는 `release_cuts:cut-b:phases:phase-2:requires:outside_cohort:phase-1`류의 block이어야 합니다.

수정 방향은 dependency-closure에서 release cohort로 인정하는 대상을 `user_approved === true` cut으로 제한하는 것입니다. `user_approved:false` cut은 후속 approved cut의 `earlierPhases`에 넣지 말고, 해당 cut 자체도 release-path closure 검증 대상에서 제외하거나 별도 tail semantics로 다뤄야 합니다. 회귀 테스트는 위처럼 false cut 뒤 true cut이 false cut phase를 require하는 케이스가 적합합니다.

확인한 테스트:
- `node --test hooks/__tests__/mpl-phase-contract-graph.test.mjs` 통과 (`44 tests / 0 failures`)
- `npm test` 통과 (`950 tests / 0 failures`)
- 별도 재현: 위 `user_approved:false` → later approved cut dependency 케이스가 현재 `valid: true`로 통과함 확인

— from codex

KyubumShin (commented):
## Review: feat(hooks) — enforce dependency-closure (B5) and released-cut immutability (D-Q6)

The Phase 1.4b PR finally closes the two open RFC items that PR #180 explicitly deferred. The split between lib-level dep-closure (write-time graph check) and hook-level immutability (state-aware diff against prior decomposition) is clean. 12 new tests cover the main paths. A few items worth surfacing before merge.

### Medium — parser-side `requires_from_phases` assumption is load-bearing but not tested for mixed shapes

`validateDependencyClosure` iterates `p.requires_from_phases` and the test comment at the bottom of the dep-closure block states:

> The validator's `requires_from_phases` extractor only records entries with an explicit `from_phase`. requires that omit `from_phase` are baseline references and must NOT produce dependency-closure errors.

But the only tests covering this contract use either an empty `requires: []` (the "baseline" path) OR a complete `from_phase` entry. There's no test for the *mixed* shape — a single phase whose `requires:` list contains both a `from_phase` entry AND a baseline-only entry. If the parser ever drifts (e.g., records both), the dep-closure check would falsely flag baseline-only requires as out-of-cohort.

Suggested test:

```js
it('extracts only entries with from_phase, ignoring baseline-only requires', () => {
  // Phase whose requires has one from_phase entry + one baseline entry.
  // Only the from_phase entry should be subject to closure check.
  // ... craft a phase block with mixed requires, mvp = [phase-2],
  // assert: mvp:phases:phase-2:requires:outside_cohort:<baseline_name> NOT emitted
});
```

This locks the parser contract that the new validator depends on.

### Medium — released-cut-absent-from-old-graph silent skip can mask state corruption

At `hooks/mpl-require-phase-contract-graph.mjs:118`:

```js
if (oldPhases === undefined) continue; // released cut absent from old graph (degenerate); cannot diff
```

If `state.release.completed_cut_ids = ['mvp']` but the prior `.mpl/mpl/decomposition.yaml` has no `mvp:` field, that's a state inconsistency (something shipped the artifact, but the source decomposition no longer carries the cut). The hook silently continues. Two readings:

- (a) **Intentional defensive** — the hook can't diff what it can't see, and surfacing this here would block routine recomposes after a graph repair. Argument: belongs in a different consistency check, not the immutability hook.
- (b) **Hidden corruption signal** — surfacing it as a distinct issue (e.g., `released_cut:<id>:missing_from_prior_graph`) would warn the operator that something dropped between `release-finalize` and now.

I lean toward (b) — the cost is one line + a clearer diagnostic message. But the current (a) behavior is defensible if accompanied by a sharper comment: "Intentionally silent: a missing cut in the old graph indicates state↔graph drift that should be diagnosed elsewhere; this hook's job is mutation-detection, not consistency-detection." Suggest the comment expansion either way.

### Minor — forward-compat shape commitment to Phase 1.6 state schema

`readCompletedCutIds` reads `state.release.completed_cut_ids` as a `string[]`. Phase 1.6 hasn't shipped the writer for this field yet, so this PR is the first commitment to the exact key path and value shape. If Phase 1.6 lands with a different shape (e.g., `state.release.completed: [{cut_id, finalized_at}]` for richer audit), this hook breaks silently (returns empty set, immutability check becomes a no-op).

Not a blocker today, but worth either:
- (a) Adding a one-line doc-comment cross-reference to the Phase 1.6 RFC section that defines this key, so the future implementer of 1.6 is forced to align with this consumer
- (b) Defining the shape now in a shared constant or types file that both this hook and the future writer import from

Either prevents the silent-no-op divergence.

### Minor — `phasesEqual` order sensitivity is correct but worth pinning with a test

```js
function phasesEqual(a, b) {
  if (a.length !== b.length) return false;
  for (let i = 0; i < a.length; i++) if (a[i] !== b[i]) return false;
  return true;
}
```

Order-sensitive, which is correct per PR #179's Rule 9 ("Never add a phase id to, remove one from, **or reorder phases within** a released cut's `.phases` list"). But no test currently covers the reorder-only case. Quick add:

```js
it('D-Q6: blocks reorder-only mutations of a released cut', () => {
  const prior = `mvp:
  derived_from: goal_contract.mvp_scope
  phases: [phase-1, phase-2]
  execution_mode: sequential
  artifact: draft_pr`;
  writeFileSync(..., graphWithMvp(prior));
  seedStateWithReleased(['mvp']);
  const reordered = graphWithMvp(`mvp:
  derived_from: goal_contract.mvp_scope
  phases: [phase-2, phase-1]
  ...`);
  const r = runHook(reordered);
  assert.equal(r.decision, 'block');
  assert.match(r.reason, /released_cut:mvp:phases:mutated/);
});
```

This pins both the order-sensitivity AND the Rule 9 ↔ hook alignment so a future "let's normalize order before diffing" refactor can't silently break the contract.

### Nit — issue-token format with phase IDs containing commas

```
released_cut:mvp:phases:mutated:old=[phase-1,phase-2]:new=[phase-1,phase-2,phase-3]
```

The `,` separator inside `[...]` is unambiguous for typical MPL phase IDs (`phase-N`, `phase-Nb`). If anyone ever lets a `,` into a phase ID, the token becomes ambiguous. Out of scope to fix here; just flagging that the format assumes a stable phase-id character class. (The existing `release_cuts:phase:overlap:...` tokens use `:` separator and aren't subject to this.)

### Verdict

Functional logic is right and the test coverage hits the important branches (cohort/forward-ref/baseline/no-mvp/pre-release/no-state.release). Suggested improvements are about *robustness* of the contract — locking the parser-side assumption with a mixed-requires test, sharpening the silent-skip semantics, and pinning the reorder-only block. None require redesign.

The two-checks-in-one-PR scope is justified: B5 + D-Q6 both depend on the same fixture refresh (the block-list parser regression test now needs an mvp so its cut's dependency resolves), and bundling them avoids landing the lib check without the hook check that complements it.

— from claude

KyubumShin (commented):
## Re-review at `8d06449` — all prior concerns addressed plus codex's high-severity finding

Verified each item:

### Fixed

- **Codex's High #1 (declined cut counted as earlier cohort)** — fixed at `mpl-phase-contract-graph.mjs:437–445` with `if (cut.user_approved !== true) continue;`. Two regression tests pin both branches of B8:
  - "rejects approved cut whose require resolves only into a DECLINED earlier cut" (declined cut's phases must NOT satisfy later approved cut's deps)
  - "skips dep-closure for declined cuts themselves (tail semantics, final phase3-gate covers)" (declined cut's own requires are tail concerns, not release-path)

  This is a real bug catch — the previous code would have let a release path silently ship with unsatisfied requires through a declined-cut phase. The comment block (`// user_approved !== true cuts are NOT release-path cohorts (RFC §10 D-Q2 / B8)`) documents the rationale tightly.

- **My Medium #1 (parser-side `requires_from_phases` mixed-shape contract)** — fixed with new test "parser: requires_from_phases extracts only entries with from_phase, ignoring baseline-only requires". The test mixes a `from_phase` entry and a baseline-only `requires:` entry on the same phase, asserts that only the `from_phase` id appears in the dep-closure issue list, AND explicitly asserts the baseline-only name does NOT appear. Locks the parser contract.

- **My Medium #2 (silent-skip semantics on released-cut-absent-from-old-graph)** — fixed at `mpl-require-phase-contract-graph.mjs:131–138` with the expanded comment that I suggested (option (a) in my prior review): "Intentionally silent on `oldPhases === undefined`: a released cut id listed in state but missing from the prior decomposition is a state↔graph drift signal that should be diagnosed by a separate consistency check, not this immutability hook. … Surfacing it here would block legitimate graph repairs after a recompose where the cut entry was already lost." Good operational reasoning; matches my (a) reading.

- **My Minor (forward-compat shape commitment to Phase 1.6)** — fixed at `mpl-require-phase-contract-graph.mjs:53–66` with a "SHAPE COMMITMENT" block: explicit `state.release.completed_cut_ids: string[]`, RFC §4.5 cross-reference, AND the forward guidance ("If Phase 1.6 needs to evolve the shape, update this consumer first or both atomically"). This is exactly the shape-pinning I asked for; the explicit "sibling field" guidance for richer audit trails is a nice addition.

- **My Minor (reorder-only block test)** — fixed with new test "D-Q6: blocks reorder-only mutations of a released cut (Rule 9 alignment)" — `[phase-1, phase-2]` → `[phase-2, phase-1]` correctly blocks. Pins both `phasesEqual` order-sensitivity AND the Rule 9 ↔ hook alignment.

Nit on issue-token format with phase IDs containing commas was correctly left unaddressed (out of scope; phase-id character class is stable in practice).

### Verdict

Strong round. The codex find on declined cuts was the real correctness gap — would have shipped a release path with unsatisfied requires in a perfectly valid-looking graph. The fix is precise (one `continue`) and the test coverage is exhaustive. My medium/minor items all landed as suggested or better.

— from claude

KyubumShin (commented):
재검토했습니다. 최신 head `8d064496e3201229692314c9c3468acde1e9cd83` 기준으로 이전 Codex blocker였던 declined cut dependency-closure 문제가 해결됐습니다.

확인한 내용:
- `user_approved !== true` cut은 release-path cohort에서 제외되어 후속 approved cut의 dependency를 만족시키지 않습니다.
- declined cut 자체의 dependency-closure는 tail semantics로 skip되어 final phase3-gate 책임과 충돌하지 않습니다.
- mixed `requires` parser 계약, released-cut reorder-only mutation, D-Q6 state shape commitment도 테스트/주석으로 고정되어 있습니다.

추가로 발견된 비중복 이슈 없습니다. LGTM.

확인한 테스트:
- `node --test hooks/__tests__/mpl-phase-contract-graph.test.mjs` 통과 (`48 tests / 0 failures`)
- `npm test` 통과 (`954 tests / 0 failures`)

— from codex

PR #187, opened 2026-05-24
feat(stage-a-1.6c-ii): release-finalize writes manifest + evidence + gate snapshot

KyubumShin:
## What

Phase 1.6c-ii of Stage A — wires `release-finalize` to write three artifacts under `.mpl/mpl/releases/{cut_id}/` before advancing the lifecycle (RFC §5.4).

| Artifact | Purpose |
|---|---|
| `release-manifest.json` | Structured manifest: `cut_id`, `phases[]`, `goal_trace.{acceptance_criteria, variation_axes}`, `artifact`, `gate_results_summary`, `pipeline_id`, `created_at` + `commit_sha`/`tree_sha`/`snapshot_ref` placeholders (1.6c-iii fills these) |
| `evidence-summary.md` | Human-readable summary: cohort header, phase status from `state.execution.phase_details`, goal trace, gate verdicts, test-agent dispatches |
| `gate-results.json` | Archival copy of `state.release.gate_results` with `archived_at` timestamp |

## Why

The 1.6c-i handler routed gate evidence into `state.release.gate_results` but ended at "PASS → release-finalize" — release-finalize itself only flipped lifecycle bookkeeping. Without the manifest write, a "released" cohort had no shippable record beyond a flag in `state.json`. 1.6c-ii closes that gap.

## Design

- **RFC** §5.4 — manifest write precedes `completed_cut_ids` append; snapshot ref (§5.4.1) and optional draft_pr/branch/tag artifacts deferred to 1.6c-iii
- **Resume brief** \`docs/roadmap/stage-a-resume-1.6c.md\` §3 (1.6c-ii scope)
- New library \`hooks/lib/mpl-release-manifest.mjs\` — pure serialization. \`resolveCutDescriptor\` resolves \`mvp\` from \`graph.mvp\` + \`contract.mvp_scope\`, extension cuts from \`graph.release_cuts[]\`
- Manifest schema is intentionally **stable** across the 1.6c-ii → 1.6c-iii boundary (snapshot fields are placeholders, not absent) so the upcoming diff is small
- Handler reads \`.mpl/goal-contract.yaml\` + \`.mpl/mpl/decomposition.yaml\`; missing descriptor → refuse to write a degraded manifest, stay at release-finalize, surface actionable message

### Invariants preserved (RFC §5.5)

- release-finalize MUST NOT set \`state.finalize_done=true\` — verified by dedicated test
- release-finalize MUST NOT transition \`current_phase\` to \`completed\` — verified by same test
- D-Q6 (PR #183): \`completed_cut_ids\` is appended ONLY after manifest write succeeds — descriptor-missing test verifies the bail path leaves \`completed_cut_ids\` untouched

## Test plan

- [x] +25 tests in this PR
  - 22 library unit tests (\`hooks/__tests__/mpl-release-manifest.test.mjs\`)
  - 5 phase-controller integration tests (3-file write, lifecycle order, §5.5 non-finalize, descriptor-missing refusal, idempotent re-write)
- [x] Full hook suite: **1019/1019 pass** (993 → 1019)
- [x] Updated 1 prior 1.6b test that asserted lifecycle advancement without seeding contract/decomposition
- [ ] Awaiting agent reviews (claude + codex)

## Agent Review Status

- [ ] from claude
- [ ] from codex

---

### DIFF
--- a/hooks/__tests__/mpl-phase-controller.test.mjs
+++ b/hooks/__tests__/mpl-phase-controller.test.mjs
@@ -1271,6 +1271,124 @@ release_cuts: []
     assert.match(out.stopReason, /cohort descriptor missing/);
   });
 
+  // PR #187 round-2 (codex + claude): bail on invalid goal-contract.
+  it('1.6c-ii (PR #187 round-2 codex+claude): release-finalize bails when contract.mvp_scope has neither AC nor AX (codex reproducer)', () => {
+    // Codex reproducer: `mvp_scope: { artifact: draft_pr }` only — the
+    // validator flags `mvp_scope.acceptance_criteria_or_variation_axes`
+    // so gcRead.valid=false. Pre-fix the handler ignored gcRead.valid
+    // and shipped a manifest with empty goal_trace, flipping D-Q6
+    // immutability for a cohort with no acceptance criteria.
+    mkdirSync(join(tmpDir, '.mpl'), { recursive: true });
+    writeFileSync(join(tmpDir, '.mpl', 'goal-contract.yaml'), `
+source:
+  runtime_goal: "x"
+  user_request_hash: "abc"
+mission:
+  goal: "g"
+  project_pivot: "pp"
+  must_ship_outcomes:
+    - "ship"
+ontology:
+  entities:
+    - foo
+variation_axes:
+  - id: AX-1
+    name: ax
+acceptance_criteria:
+  - id: AC-1
+    statement: "ac"
+e2e_policy:
+  real_runtime_required: true
+  mock_allowed: false
+  placeholder_assertions_allowed: false
+security_policy:
+  required: false
+completion_evidence:
+  required_artifacts:
+    - .mpl/mpl/audit-report.json
+  require_commit: false
+  require_finalize_timestamps: true
+mvp_scope:
+  artifact: draft_pr
+`);
+    writeDecompositionWithMvp(tmpDir);
+    const out = runStopHook(tmpDir, {
+      current_phase: 'release-finalize',
+      release: {
+        current_cut_id: 'mvp', completed_cut_ids: [], fix_loop_count: 0,
+        pending_artifact: null, max_fix_loops: 3,
+        gate_results: {
+          hard1_baseline: { exit_code: 0 }, hard2_coverage: { exit_code: 0 }, hard3_resilience: { exit_code: 0 },
+        },
+      },
+    });
+    const state = readState();
+    // No advancement.
+    assert.strictEqual(state.current_phase, 'release-finalize');
+    assert.strictEqual(state.release.current_cut_id, 'mvp');
+    assert.deepStrictEqual(state.release.completed_cut_ids, []);
+    assert.match(out.stopReason, /goal-contract is invalid/);
+    assert.match(out.stopReason, /mvp_scope\.acceptance_criteria_or_variation_axes/);
+    // No file written.
+    assert.equal(existsSync(join(tmpDir, '.mpl', 'mpl', 'releases', 'mvp', 'release-manifest.json')), false);
+  });
+
+  it('1.6c-ii (PR #187 round-2 codex+claude): release-finalize bails when contract validator flags any structural failure (defense-in-depth)', () => {
+    // Any contract validity failure must block release-finalize — even
+    // unrelated-looking ones (e.g., unknown AC id) — because a broken
+    // contract at release-time means goal_trace cannot be trusted.
+    mkdirSync(join(tmpDir, '.mpl'), { recursive: true });
+    writeFileSync(join(tmpDir, '.mpl', 'goal-contract.yaml'), `
+source:
+  runtime_goal: "x"
+  user_request_hash: "abc"
+mission:
+  goal: "g"
+  project_pivot: "pp"
+  must_ship_outcomes:
+    - "ship"
+ontology:
+  entities:
+    - foo
+variation_axes:
+  - id: AX-1
+    name: ax
+acceptance_criteria:
+  - id: AC-1
+    statement: "ac"
+e2e_policy:
+  real_runtime_required: true
+  mock_allowed: false
+  placeholder_assertions_allowed: false
+security_policy:
+  required: false
+completion_evidence:
+  required_artifacts:
+    - .mpl/mpl/audit-report.json
+  require_commit: false
+  require_finalize_timestamps: true
+mvp_scope:
+  acceptance_criteria: [AC-DOES-NOT-EXIST]
+  variation_axes: [AX-1]
+  artifact: draft_pr
+`);
+    writeDecompositionWithMvp(tmpDir);
+    const out = runStopHook(tmpDir, {
+      current_phase: 'release-finalize',
+      release: {
+        current_cut_id: 'mvp', completed_cut_ids: [], fix_loop_count: 0,
+        pending_artifact: null, max_fix_loops: 3,
+        gate_results: {
+          hard1_baseline: { exit_code: 0 }, hard2_coverage: { exit_code: 0 }, hard3_resilience: { exit_code: 0 },
+        },
+      },
+    });
+    const state = readState();
+    assert.strictEqual(state.current_phase, 'release-finalize');
+    assert.deepStrictEqual(state.release.completed_cut_ids, []);
+    assert.match(out.stopReason, /goal-contract is invalid/);
+  });
+
   it('1.6c-ii (PR #187 codex review): release-finalize refuses to advance when graph.mvp.phases is empty', () => {
     // Decomposer wrote `mvp:` block but `phases: []` (mechanical mapping
     // yielded no membership). Manifest would assert "released" with no
--- a/hooks/__tests__/mpl-phase-controller.test.mjs
+++ b/hooks/__tests__/mpl-phase-controller.test.mjs
@@ -1,7 +1,7 @@
 import { describe, it, beforeEach, afterEach } from 'node:test';
 import assert from 'node:assert/strict';
 import { execFileSync } from 'node:child_process';
-import { mkdtempSync, rmSync, mkdirSync, writeFileSync, readFileSync } from 'fs';
+import { mkdtempSync, rmSync, mkdirSync, writeFileSync, readFileSync, existsSync } from 'fs';
 import { join, dirname } from 'path';
 import { fileURLToPath } from 'url';
 import { tmpdir } from 'os';
@@ -730,10 +730,38 @@ mvp_scope:
   });
 
   it('1.6b: release-finalize appends current cohort to completed_cut_ids and clears current_cut_id', () => {
-    mkdirSync(join(tmpDir, '.mpl'), { recursive: true });
+    // After 1.6c-ii, release-finalize requires goal-contract + decomposition
+    // to build the release-manifest before advancing the lifecycle. Seed
+    // both so the original 1.6b lifecycle assertion still exercises the
+    // append+clear+route path. The helpers `writeGoalContractWithMvp` and
+    // `writeDecompositionWithMvp` are defined in the 1.6c-ii block below;
+    // we inline equivalents here to avoid a forward-declaration dependency.
+    mkdirSync(join(tmpDir, '.mpl', 'mpl'), { recursive: true });
+    writeGoalContractWithMvp(tmpDir);
+    writeFileSync(join(tmpDir, '.mpl', 'mpl', 'decomposition.yaml'), `
+graph_version: "1.0"
+generated_by: mpl-decomposer
+recompose_count: 0
+completed_phase_policy: immutable
+execution_tiers:
+  - tier: 1
+    phases: [phase-1, phase-2]
+mvp:
+  phases: [phase-1, phase-2]
+  execution_mode: sequential
+  artifact: draft_pr
+  derived_from: mvp_scope
+release_cuts: []
+`);
     const out = runStopHook(tmpDir, {
       current_phase: 'release-finalize',
-      release: { current_cut_id: 'mvp', completed_cut_ids: [], fix_loop_count: 0, pending_artifact: null },
+      release: {
+        current_cut_id: 'mvp', completed_cut_ids: [], fix_loop_count: 0, pending_artifact: null,
+        max_fix_loops: 3,
+        gate_results: {
+          hard1_baseline: { exit_code: 0 }, hard2_coverage: { exit_code: 0 }, hard3_resilience: { exit_code: 0 },
+        },
+      },
     });
     const state = readState();
     assert.deepStrictEqual(state.release.completed_cut_ids, ['mvp']);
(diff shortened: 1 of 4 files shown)


KyubumShin (commented):
Solid implementation — pure-library / handler-owns-IO split, RFC §5.5 invariants explicitly tested, idempotent re-write covered, descriptor-missing path refuses to advance. 1019/1019 hook tests pass. Below are non-blocking concerns worth a thought before 1.6c-iii lands on top.

### 1. File mode `0o600` for manifest files — intentional?

`hooks/mpl-phase-controller.mjs:919,924,929` writes all three artifacts with `{ mode: 0o600 }`. That's owner-read-write only. For `gate-results.json` (archival snapshot) this is defensible, but `evidence-summary.md` is explicitly framed as a human-skimmable artifact, and `release-manifest.json` is the structured record other tools (CI, `mpl-status`, future `mpl-release-list`) will want to read. On a multi-user dev box or shared CI runner with a process running as a different uid, 0o600 will silently break consumers.

Is the intent to treat these as "private" releases? If so, fine — but consider documenting it on `RELEASE_DIR_REL_PATH`. Otherwise `0o644` matches conventions elsewhere in the repo (`mpl-state` writes don't pin mode).

### 2. 3-file write is non-atomic — partial-failure window

`mpl-phase-controller.mjs:914-941`: if `release-manifest.json` writes successfully and `evidence-summary.md` then fails (disk full, permission flip, etc.), the catch surfaces the error and bails before `completed_cut_ids` append — good. But the on-disk state is now `{ manifest exists, summary missing, gate-snapshot missing }`. The idempotency test (`mpl-phase-controller.test.mjs:207-236`) covers re-entry overwriting a stale manifest, but doesn't cover this asymmetric-partial case.

The simple fix is temp-write + rename per file (`writeFileSync(tmp); renameSync(tmp, final)`) so each file is atomically present-or-absent. Probably defer to 1.6c-iii since 1.6c-iii adds git operations that have the same "what if it fails mid-way" question and the two should be designed together.

### 3. `buildGateResultsSnapshot` clone is shallow

`hooks/lib/mpl-release-manifest.mjs:706-709`:
```js
const clone = (entry) => {
  if (!entry || typeof entry !== 'object') return null;
  return { ...entry };
};
```

The defensive-clone test at `mpl-release-manifest.test.mjs:437-445` only verifies top-level mutation isolation. If a gate entry ever grows a nested object (e.g., `{ exit_code: 0, details: { ... } }`), mutating that nested object on the source would leak into the snapshot. Today every gate-result entry is flat, so this is theoretical, but a `structuredClone(entry)` is a one-line upgrade with no behavior change for current shapes.

### 4. Extension-cut AC/AX is silently empty — Stage B blind spot

`hooks/lib/mpl-release-manifest.mjs:621-627`: extension cuts get hardcoded `acceptance_criteria: []` and `variation_axes: []` with the inline comment "Extension cuts don't carry AC/AX in Stage A — Future Stage B may extend." When Stage B does extend, `goal_trace` for non-mvp cohorts will silently report empty lists rather than the actual AC/AX. A future reader updating release-cut schema may not think to revisit this resolver.

Two cheap defenses: (a) tighten the comment to "if Stage B adds release_cuts[].acceptance_criteria, propagate it here", or (b) already read `found.acceptance_criteria ?? []` so the resolver auto-picks-up the field if it appears. Prefer (b) — it's literally one character per array, and a future schema migration won't need a code change here.

### 5. Graph read is inlined; goal-contract uses a helper — asymmetric

`mpl-phase-controller.mjs:686-696`: `readGoalContract(cwd)` is a clean helper but the graph read is manual `existsSync` + `readFileSync` + try-catch. Minor — a `readPhaseContractGraph(cwd)` wrapper would match the contract-read pattern and remove ~10 lines from the handler. Worth doing if you touch this file again for 1.6c-iii.

### 6. Updated 1.6b test inlines decomposition YAML

`mpl-phase-controller.test.mjs:25-41`: the existing 1.6b test was updated to seed contract+decomposition inline rather than reusing `writeGoalContractWithMvp` / `writeDecompositionWithMvp`, with a comment "we inline equivalents here to avoid a forward-declaration dependency." Two copies of the same YAML now exist in the same file (the second is at line 62-78). If decomposition schema evolves, both must update in lockstep or one test silently passes against a stale shape.

Hoisting the helper functions to top-of-describe-block would let the 1.6b test reuse them. Not a blocker, but the comment effectively admits it's a workaround.

---

None of the above is a merge blocker — manifest contract, lifecycle invariants, and idempotency are all correct and tested. (1) and (2) are the most worth addressing; (3)/(4)/(5)/(6) are quality-of-life.

— from claude

KyubumShin (commented):
리뷰 코멘트입니다.

1. **High**: `release-finalize`가 계약/분해 산출물 중 하나가 빠진 상태에서도 degraded manifest를 쓰고 cohort를 완료 처리할 수 있습니다. 핸들러 주석과 PR 설명은 goal-contract와 decomposition이 모두 필요하고, 누락 시 `completed_cut_ids` append를 하지 않아야 한다고 되어 있습니다 (`hooks/mpl-phase-controller.mjs:689`). 그런데 `resolveCutDescriptor()`는 `mvp`에서 `graph.mvp` 또는 `contract.mvp_scope` 중 하나만 있어도 descriptor를 반환합니다 (`hooks/lib/mpl-release-manifest.mjs:41`), 그래서 `manifest`가 null이 아니게 됩니다. 재현상 `.mpl/goal-contract.yaml`만 있고 `.mpl/mpl/decomposition.yaml`이 없는 `release-finalize` 상태에서 Stop hook이 `release-manifest.json`을 `phases: []`로 쓰고, `completed_cut_ids:["mvp"]`, `current_cut_id:null`, `current_phase:"phase3-gate"`로 전진합니다. 이러면 RFC §5.4의 “manifest write precedes release completion”이 사실상 빈 phase manifest로 만족되고, D-Q6 released-cut immutability가 실제 cut phase 목록 없이 켜집니다. `mvp` descriptor는 최소한 `contract.mvp_scope`와 non-empty `graph.mvp.phases`를 모두 요구해야 하고, contract-only / graph-only 누락 케이스가 lifecycle을 전진하지 않는 회귀 테스트가 필요합니다.

확인:
- `node --test hooks/__tests__/mpl-release-manifest.test.mjs hooks/__tests__/mpl-phase-controller.test.mjs` 통과 (`90 tests / 0 failures`)
- `npm test` 통과 (`1019 tests / 0 failures`)

— from codex

KyubumShin (commented):
Re-review at `b28cdd0`. Every concern I raised is either fixed in this commit or explicitly deferred to 1.6c-iii with matching rationale. Codex's strict-descriptor bug (the High) is the right call — the OR-logic was a real silent-degradation path and the new AND + empty-phases guards close it cleanly. +4 strict-descriptor regression tests in the library and +3 integration tests at the handler level give the fix proper coverage.

Spot-checks on the fixes I asked for:

- **#1 mode 0o644**: applied to all three artifacts with a rationale comment that contrasts against state.json's 0o600. The `0o644` integration test guards against future mode regression (with a sensible win32 skip).
- **#3 structuredClone**: drop-in replacement; the nested-mutation test pins behavior. Requires Node 17+ but the codebase already uses post-Node-17 features so that's fine.
- **#4 extension-cut AC/AX**: the `Array.isArray(found.acceptance_criteria) ? [...found.acceptance_criteria] : []` shape is exactly what Stage B can extend into without code change. Test pins the forward-compat.
- **#6 hoisted helper reuse**: `writeDecompositionWithMvp` is reused in the 1.6b test via function-declaration hoisting. The new inline comment explains the hoisting rather than just removing it — good footnote for the next reader.

Deferred items match my recommendation:
- **#2 atomic write**: rightly bundled with 1.6c-iii git ops (they share the partial-failure question)
- **#5 readPhaseContractGraph wrapper**: 1.6c-iii will read snapshot-ref state too, natural place to refactor

### One forward-looking thought (non-blocking)

The strict check at `mpl-release-manifest.mjs:52` is `if (!mvpGraph || !mvpScope) return null;` — so a truthy-but-empty `mvp_scope: {}` (or `{ acceptance_criteria: null, variation_axes: null }`) still passes the gate. It then drops through the `Array.isArray(...) ? [...] : []` fallback at lines 60-65, which produces a manifest with `goal_trace: { acceptance_criteria: [], variation_axes: [] }`. That's exactly the kind of "asserts released, ships nothing" shape the empty-phases guard was added to prevent — but on the goal_trace axis instead of the phases axis.

Probably the right line for Stage A — early bootstrap pipelines may legitimately have no declared AC/AX while still wanting to test the release pipeline end-to-end. But worth a moment of thought: if Stage B treats goal_trace as load-bearing, an analogous "at least one AC OR at least one AX" guard would close the symmetric hole. Not a 1.6c-ii blocker — file under "consider when AC/AX become mandatory."

### Verdict

1029/1029 hook tests pass, manifest contract holds across the change, RFC §5.5 invariants still tested. The descriptor narrowing is sound and well-tested. No new concerns from this pass — LGTM at `b28cdd0`.

— from claude

KyubumShin (commented):
재검토했습니다. 이전에 남긴 contract-only/decomposition-missing High는 `b28cdd0`에서 해결됐습니다. contract-only, graph-only, empty `graph.mvp.phases` 모두 `release-finalize`에 머물도록 바뀐 점 확인했습니다.

1. **High**: `release-finalize`가 invalid goal contract를 무시하고 release를 완료할 수 있습니다. `readGoalContract()`는 빈/잘못된 `mvp_scope`를 `valid:false`로 판정하지만, handler는 `gcRead.valid`를 보지 않고 `gcRead.exists ? gcRead.contract : null`만 넘깁니다 (`hooks/mpl-phase-controller.mjs:693`). 그래서 validator가 이미 `mvp_scope.acceptance_criteria_or_variation_axes`로 invalid 처리하는 계약도, decomposition에 `mvp.phases`만 있으면 `buildReleaseManifest()`가 manifest를 만들고 `completed_cut_ids:["mvp"]`, `current_cut_id:null`, `current_phase:"phase3-gate"`까지 전진합니다. 재현상 `mvp_scope: { artifact: draft_pr }`만 있는 계약에서 `release-manifest.json`이 `goal_trace: { acceptance_criteria: [], variation_axes: [] }`로 기록됐습니다. 이건 이번 strict descriptor fix의 의도와 같은 축의 문제입니다. `release-finalize`는 `!gcRead.valid`를 descriptor missing/fault path로 취급하거나, 최소한 `resolveCutDescriptor("mvp")`가 non-empty AC/AX와 supported artifact를 요구해야 합니다. 특히 goal contract가 decomposition 이후 drift/부분수정된 경우, released cut immutability가 빈 goal trace를 기준으로 켜질 수 있습니다.

확인:
- 기존 High 재현: 이제 manifest를 쓰지 않고 `release-finalize`에 머묾
- invalid `mvp_scope` 재현: 현재는 빈 `goal_trace` manifest를 쓰고 release 완료 처리됨
- `node --test hooks/__tests__/mpl-release-manifest.test.mjs hooks/__tests__/mpl-phase-controller.test.mjs` 통과 (`100 tests / 0 failures`)
- `npm test` 통과 (`1029 tests / 0 failures`)

— from codex

KyubumShin (commented):
Retracting my prior LGTM at `b28cdd0` in light of codex's re-review.

Codex's new High — `release-finalize` writes a manifest with empty `goal_trace` for a contract whose `mvp_scope` is missing AC/AX — is the concrete realization of the forward-looking thought I parked at the end of my last review ("a truthy-but-empty `mvp_scope: {}` … still passes the strict gate and produces `goal_trace: { acceptance_criteria: [], variation_axes: [] }`"). Codex's reproducer (`mvp_scope: { artifact: draft_pr }` only) plus the `gcRead.valid` ignored at `hooks/mpl-phase-controller.mjs:693` makes this a real D-Q6 hazard, not a hypothetical:

- `readGoalContract()` already determines `valid: false` for these contracts (the validator flags `mvp_scope.acceptance_criteria_or_variation_axes`)
- handler reads only `gcRead.exists`, drops `gcRead.valid` on the floor
- `resolveCutDescriptor` only checks truthy presence of `mvp_scope`, not its content
- net: a degraded "released" manifest ships, `completed_cut_ids` flips immutable for a cohort with no goal trace

This is symmetric to the empty-phases guard that the codex-High strict-descriptor fix added on the `phases` axis. The two should be fixed together because they're the same class of bug ("a cohort declared 'released' with no real content").

### Verdict

I withdraw my LGTM. Concur with codex's High — treat as a 1.6c-ii blocker, not a 1.6c-iii deferral. Two options for the fix:

1. **Handler layer**: at `hooks/mpl-phase-controller.mjs:693`, route `!gcRead.valid` through the same "cohort descriptor missing" bail path that the contract/decomposition-missing case takes. Smallest diff; reuses existing tested bail mechanism.
2. **Library layer**: tighten `resolveCutDescriptor`'s mvp branch to also reject when both `acceptance_criteria` and `variation_axes` come out empty. More defensible against future callers that bypass `readGoalContract`'s validator.

Option 1 alone fixes the codex reproducer; option 2 also closes the case where a caller hand-builds a contract object bypassing the validator. Probably both, with option 1 as the immediate guard and option 2 as the defense-in-depth — and a regression test that mirrors codex's reproducer (`mvp_scope: { artifact: 'draft_pr' }` only → stays at `release-finalize`, no `completed_cut_ids` append).

— from claude

KyubumShin (commented):
재검토했습니다. 제가 이전에 남긴 invalid `mvp_scope` High는 `686bdfc`에서 해결됐습니다.

확인한 내용:
- `release-finalize`가 `readGoalContract()` 결과의 `valid:false`를 확인하고, invalid goal contract에서는 manifest를 쓰지 않고 `release-finalize`에 머뭅니다.
- 제 재현 케이스인 `mvp_scope: { artifact: draft_pr }`만 있는 계약은 이제 `mvp_scope.acceptance_criteria_or_variation_axes`를 stopReason에 표시하고, `completed_cut_ids`를 append하지 않습니다.
- `resolveCutDescriptor()`도 AC/AX가 모두 비어 있으면 `null`을 반환하도록 library-layer 방어가 추가되어, handler를 우회하는 호출자도 빈 `goal_trace` manifest를 만들 수 없습니다.
- contract-only/decomposition-missing, graph-only, empty `graph.mvp.phases` 회귀 방어도 유지됩니다.

테스트:
- invalid `mvp_scope` Stop hook 재현: manifest 미작성, `current_phase:"release-finalize"`, `completed_cut_ids:[]`
- `node --test hooks/__tests__/mpl-release-manifest.test.mjs hooks/__tests__/mpl-phase-controller.test.mjs` 통과 (`105 tests / 0 failures`)
- `npm test` 통과 (`1034 tests / 0 failures`)

추가 blocker 없습니다. LGTM at `686bdfc`.

— from codex

KyubumShin (commented):
Re-review at `686bdfc`. Round-2 fix matches the two-layer approach codex and I jointly recommended:

- **Handler** (`mpl-phase-controller.mjs:693-713`): `gcRead.exists && !gcRead.valid` now routes through a dedicated bail with the validator's `gcRead.missing` joined into `stopReason` — actionable, points the user at `validateGoalContractText` as the source of truth.
- **Library** (`mpl-release-manifest.mjs:62-76`): `if (acceptance.length === 0 && axes.length === 0) return null;` — symmetric to the round-1 empty-phases guard, matches the validator's `_or_` semantics (rejects only when BOTH empty; AC-only or AX-only contracts still accepted).

Test coverage is clean:
- Codex's reproducer (`mvp_scope: { artifact: draft_pr }` only) → bails with `goal-contract is invalid (mvp_scope.acceptance_criteria_or_variation_axes)` and no manifest written
- Defense-in-depth case (unknown AC id) → bails on the same `goal-contract is invalid` path, proving the handler catches any validator failure, not just AC+AX-empty
- Library unit tests pin the `_or_` semantics: AC-only OK, AX-only OK, BOTH-empty rejected

One minor observation, not a blocker:

- The handler now validates goal-contract but not decomposition — a malformed-but-parseable `decomposition.yaml` (e.g., `mvp.phases: [bad-id]` referencing a non-existent phase) would still pass the parse-and-then-strict checks because there's no decomposer-side validator equivalent. Probably the right Stage A scope (decomposition is decomposer-mechanically produced, not user-edited), but worth noting when Stage B opens up user-editable extension cuts.

Verdict: LGTM at `686bdfc`. 1034/1034 hook tests pass, both axes (phases-empty + AC+AX-empty) now have symmetric guards, validator integration is correct.

— from claude

PR #186, opened 2026-05-24
feat(stage-a-1.6c-i): release-gate scoped Hard 1/2/3 evidence routing

KyubumShin:
## What

Phase 1.6c-i of Stage A — wires the `release-gate` handler to consume scoped gate evidence from `state.release.gate_results` and route per RFC §5.3.1.

- **Schema v5 → v6** with additive migration adding:
  - `state.release.gate_results` — parallel to top-level `state.gate_results`, populated by mpl-gate-recorder when `current_phase == 'release-gate'` only (RFC §5.5 isolation)
  - `state.release.max_fix_loops` — scoped retry budget (default **3**, separate from top-level default 10)
- **`release-gate` handler** (replaces 1.6b stub):
  - PASS → release-finalize
  - FAIL within budget → increment scoped `fix_loop_count`, route back to phase2-sprint preserving `current_cut_id`, reset scoped evidence
  - FAIL at threshold → circuit-break: stay at release-gate, pin cohort, surface ⛔ with user intervention guidance
  - MISSING → continue with stopReason; honors workspace strict mode (block) same as phase3-gate
- **Reuses `checkGateResults`** via a one-line adapter — structured-vs-legacy decision tree stays single-sourced across both gate paths.

## Why

The 1.6b stub immediately passed through to release-finalize regardless of evidence, meaning every cohort shipped without verification. 1.6c-i adds the consumer side of the scoped Hard 1/2/3 gate. Evidence *production* (running scoped commands, deciding Hard 2 affected scope, gate-recorder routing) is intentionally deferred to a follow-up — splitting the surfaces keeps each PR review-friendly per the 1.6c sub-split plan in \`docs/roadmap/stage-a-resume-1.6c.md\`.

## Design

- RFC: \`docs/roadmap/stage-a-mvp-cuts-rfc.md\` §5.3.1 (failure path), §5.5 (whole-pipeline isolation)
- Resume brief: \`docs/roadmap/stage-a-resume-1.6c.md\` §3 (1.6c-i scope), §7 (open questions Q1, Q3 decided here)

### Open question decisions

- **Q1 SSOT**: state.json (state.release.gate_results) — read-side parity with state.gate_results; write-side file mirror lands in 1.6c-ii
- **Q3 max_fix_loops**: separate state.release.max_fix_loops field with default 3 (per RFC §5.3.1)
- **Q2 (gate-recorder routing) / Q4 (affected-tests resolution)**: deferred — this PR is consumer-only

### Invariants preserved

[...]

PR #185, opened 2026-05-24
feat(orchestrator): wire Stage A release path — sprint routing + state handlers (1.6b)

KyubumShin:
## What

Stage A implementation **phase 1.6b**: orchestrator state-machine wiring that consumes the foundation from PR #184 (state.release subtree) and PR #182 (decomposer mvp derivation).

- `hooks/mpl-phase-controller.mjs` (+82, -5): release-gate/release-finalize handlers, phase2-sprint lifecycle init + cohort-aware completion routing
- `hooks/__tests__/mpl-phase-controller.test.mjs` (+165): 9 new tests covering init, routing both directions, both handlers, idempotency, defensive paths

Full hook suite: **971/971 pass** (+9, 0 regression).

## Why — End-to-End Stage A Flow Activated

After this PR, the full Stage A loop runs end-to-end for single-cohort (mvp-only) projects:

```
user declares mvp_scope in interview      (#181)
  → goal-contract.yaml carries mvp_scope  (#178 parser)
  → decomposer derives decomposition.mvp  (#182)
  → THIS PR: sprint inits current_cut_id="mvp"
  → THIS PR: sprint completion → release-gate (cohort active)
  → release-gate stub → release-finalize
  → THIS PR: release-finalize appends mvp to completed_cut_ids
            → clears current_cut_id → phase3-gate
  → D-Q6 immutability hook (#183) now ACTIVE for recompose
```

Before this PR, mvp_scope was parseable + validated + derived but never *routed on*. The release path existed only as state schema (PR #184). This PR is the wiring that makes it actually run.

## Design

- Source: `docs/roadmap/stage-a-mvp-cuts-rfc.md` §5.1 (release path states), §5.4.2 (cut activation), §4.5 (state.release lifecycle), §10 D-Q2 (auto-proposed cuts are Stage B).
- Companion PRs (all merged): #178, #179, #180, #181, #182, #183, #184.
- **Stage A simplification**: decomposer emits `release_cuts: []` (PR #182 per RFC §10 D-Q2), so release-finalize never advances to a next cut — always routes to phase3-gate. Multi-cohort chaining is Stage B work and would require cohort-aware PLAN.md filtering (also Stage B).
- **Lifecycle init location**: lazily inside phase2-sprint case rather than at mpl-decompose. Reason: mpl-decompose case doesn't write current_phase=phase2-sprint itself (orchestrator/commands do). Putting the init at sprint entry makes it robust regardless of who set the phase.

## Risk

[...]

PR #188, opened 2026-05-24
feat(stage-a-1.6c-iii): snapshot ref + artifact creation in release-finalize

KyubumShin:
## What

Phase 1.6c-iii — final Stage A phase. Fills the snapshot-ref placeholders 1.6c-ii left and adds the optional user-visible artifact (`tag` / `branch` / `draft_pr`) per RFC §5.4.1.

### Snapshot ref

At the start of `release-finalize`, before manifest serialize:
- `git rev-parse HEAD` → `commit_sha`
- `git rev-parse HEAD^{tree}` → `tree_sha`
- `git update-ref refs/mpl/releases/{cut_id} <commit>` → `snapshot_ref`

These three fields populate the manifest. On any git failure (not a repo, etc.) the manifest is still written with null placeholders and `artifact_creation_failed = {type: 'snapshot_ref', reason}`. Lifecycle still advances (RFC §5.4 best-effort post-step).

### Artifact creation

| Type | Action |
|---|---|
| `release_manifest` | no-op (snapshot + manifest only) |
| `tag` | `git tag mpl-release-{cut_id}` + push to origin |
| `branch` | `git branch mpl/release/{cut_id}` + push to origin |
| `draft_pr` | branch + push prerequisite, then `gh pr create --draft` with body linking the snapshot ref + "do not push to this branch" warning |

Library NEVER throws — every helper returns `{ok: true, ...}` or `{ok: false, reason}`. Soft-fail surface for missing remote + missing `gh` CLI. Local refs survive push failures for manual retry.

### Picked up from PR #187 deferred items

- **claude #2 (atomic write)**: all 3 release artifacts now write through temp+rename helper. Mid-write failure leaves prior generation intact instead of partial-mix.
- **claude #5 (graph helper)**: not extracted — would not reduce surface meaningfully; leaving inline read+parse.

### Failure surfacing

`release-finalize` stopReason now ends with a status-aware tail:
- `snapshot refs/mpl/releases/mvp @ <sha12>. artifact=tag created.`
- `⚠ snapshot ref FAILED (...) — artifact_creation_failed recorded. artifact=tag skipped (snapshot ref failed).`
- `snapshot refs/mpl/releases/mvp @ <sha12>. ⚠ artifact=tag FAILED (...).`

## Why

Closes Stage A. After this PR, a single-cohort MVP project:
1. opts into MVP via Stage 1.4.5 checklist (PR #181)
2. runs release-gate scoped Hard 1/2/3 (PR #186)
3. ships at release-finalize with manifest + evidence + gate snapshot (PR #187)
4. **gets a snapshot-pinned commit + optional user-visible artifact (this PR)**

## Design

- RFC: \`docs/roadmap/stage-a-mvp-cuts-rfc.md\` §5.4.1
- Resume brief: \`docs/roadmap/stage-a-resume-1.6c.md\` §3 (1.6c-iii scope)

## Test plan

- [x] +21 tests in this PR (16 library unit + 5 phase-controller integration)
- [x] Full hook suite: **1055/1055 pass** (1034 → 1055)
- [x] Atomic write verified by no-.tmp-residue assertion
- [x] gh stub injected via PATH override for draft_pr success path
- [ ] Awaiting agent reviews (claude + codex)

## Agent Review Status

- [ ] from claude
- [ ] from codex

---

### DIFF
--- a/hooks/__tests__/mpl-phase-controller.test.mjs
+++ b/hooks/__tests__/mpl-phase-controller.test.mjs
@@ -1,7 +1,7 @@
 import { describe, it, beforeEach, afterEach } from 'node:test';
 import assert from 'node:assert/strict';
 import { execFileSync } from 'node:child_process';
-import { mkdtempSync, rmSync, mkdirSync, writeFileSync, readFileSync, existsSync, statSync } from 'fs';
+import { mkdtempSync, rmSync, mkdirSync, writeFileSync, readFileSync, existsSync, statSync, readdirSync } from 'fs';
 import { join, dirname } from 'path';
 import { fileURLToPath } from 'url';
 import { tmpdir } from 'os';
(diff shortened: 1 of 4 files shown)


KyubumShin (commented):
Solid library shape — pure helpers, never-throws contract, soft-fail with `{ok, reason}`, atomic temp+rename for the 3-file write, gh stub injected via PATH for the success path test. RFC §5.4 best-effort post-step semantics preserved. 1055/1055 hook tests pass.

Three concerns worth resolving before merge, plus one forward-looking thought:

### 1. Tag/branch/draft_pr break release-finalize idempotency

PR #185 established release-finalize idempotency at the lifecycle level (re-entering with a cohort already in `completed_cut_ids` was a clean no-op). PR #187 added the manifest-overwrite test (`mpl-phase-controller.test.mjs:207-236`) to extend that contract to the manifest write. This PR adds the snapshot ref using `git update-ref` (idempotent — re-pointing the ref is allowed, and the `createSnapshotRef` test at `mpl-release-artifact.test.mjs:303-321` verifies this) — but breaks idempotency for the tag/branch/draft_pr paths:

- `hooks/lib/mpl-release-artifact.mjs:589` — `git tag mpl-release-mvp <sha>` fails with "tag already exists" on the second invocation. The current code surfaces this as `git tag failed: ...` and routes through `artifact_creation_failed.type='tag'`.
- `hooks/lib/mpl-release-artifact.mjs:615` — same for `git branch mpl/release/mvp <sha>`.
- For `draft_pr`, the cascading branch creation hits the same wall before the `gh pr create` step runs.

Concretely: if release-finalize succeeds, then fails partway through a later cohort-rebuild and re-enters, the artifact step now wedges with a spurious failure even though the prior outcome was correct. There's no test in this PR that exercises this re-entry path — the existing idempotency test uses the default `draft_pr` against a no-remote fixture where the branch push fails on the first attempt anyway, masking the issue.

Two reasonable fixes:
- (a) Check ref existence first; if the existing ref points at the same `commitSha`, return `{ok: true, ...}` with a `noop: true` field. (b) Use `git tag -f` / `git branch -f` to match the snapshot-ref `update-ref` semantics. (a) is closer to "honest about state", (b) is the one-line change.

Add a regression test that runs `release-finalize` twice with `artifact: tag` and asserts the second run still reports `artifact=tag created` (or `noop`) rather than a spurious failure.

### 2. Manifest re-write on artifact failure silently swallows write errors

`hooks/mpl-phase-controller.mjs:870-877`:
```js
try {
  atomicWriteFile(manifestPath, JSON.stringify(manifest, null, 2) + '\n', 0o644);
} catch {
  // Manifest re-write failure is itself non-fatal — the original manifest...
}
```

The comment explains the design choice (lifecycle still advances per RFC §5.4), but the result is that the persisted `release-manifest.json` can claim `artifact_creation_failed: null` while the user's stopReason says `⚠ artifact=tag FAILED (...)`. State on disk diverges from the user-visible truth, and a downstream consumer reading the manifest later sees no failure record.

Two cheap mitigations: (a) `console.log` a `WARN` line to stderr noting the manifest re-write failed and recommending manual inspection — at least the user has a breadcrumb. (b) Include the artifact failure in the FIRST manifest write by attempting artifact creation before the manifest write, not after. Right now ordering is `manifest-write → artifact-attempt → manifest-rewrite`; swap to `artifact-attempt → manifest-write` and the re-write isn't needed. Snapshot ref is already captured pre-manifest-write — extending the same pattern to user-visible artifacts is consistent.

### 3. Snapshot ref is local-only — never pushed

`createSnapshotRef` runs `git update-ref refs/mpl/releases/{cutId}` but doesn't push that ref. The tag and branch artifacts push their refs to origin; the snapshot ref (which is the canonical pin for "this manifest was built from THIS commit") stays local. If the developer's machine dies or a fresh clone happens, the snapshot ref is gone — only the manifest's `commit_sha` string remains, which a consumer would have to manually verify against a separately-pushed ref.

Two reads of the design:
- **Intentional**: the manifest file is the canonical record (it's checked in / shipped), and the local ref is just for the user's `git rev-parse refs/mpl/releases/mvp` workflow. In that case, document the limitation in the lib comment so future readers don't assume the ref is durable across machines.
- **Oversight**: the snapshot ref should be pushed alongside tag/branch so downstream CI can `git fetch origin 'refs/mpl/releases/*:refs/mpl/releases/*'` and verify the manifest. In that case, add a `tryPushRef(cwd, snapshot.snapshot_ref)` call after `createSnapshotRef` succeeds, soft-failing the same way tag/branch do.

I lean toward the latter — the cost is small and it makes "released state immutability" verifiable from outside the originating machine. But this is a design call, not a bug.

### 4. Forward-looking: HEAD-may-drift-past-cohort-tip

`createSnapshotRef` pins HEAD at release-finalize time. If the user makes commits between release-gate PASS and release-finalize firing (debug work, unrelated commits), the snapshot pins those, not the cohort's true terminal commit. The pipeline doesn't currently enforce a clean tree at release-finalize entry.

Probably right for Stage A scope (release-gate → release-finalize is a tight loop in the normal flow), but worth a comment in the snapshot ref helper noting that callers are responsible for tree-state hygiene. When Stage B / multi-cohort flows land, a "release-finalize must run on the cohort tip" invariant probably becomes load-bearing.

---

(1) and (2) are the most material. (3) is a design question that should be answered before merge even if the answer is "intentional, document it." (4) is a comment-quality nit.

— from claude

KyubumShin (commented):
리뷰 코멘트입니다. Claude가 이미 tag/branch idempotency, artifact failure re-write, snapshot ref push 여부를 짚어서 그 내용은 반복하지 않겠습니다. 별도 blocker 하나 확인했습니다.

1. **High**: snapshot ref 생성 실패가 release 완료 조건을 우회합니다. RFC §5.4는 release manifest가 `commit_sha`, `tree_sha`, `snapshot_ref`를 포함해야 하고, `completed_cut_ids` append 조건을 “gate PASS + release-manifest write + snapshot ref creation all succeed”로 둡니다 (`docs/roadmap/stage-a-mvp-cuts-rfc.md:229`, `docs/roadmap/stage-a-mvp-cuts-rfc.md:232`). 그런데 현재 handler는 `createSnapshotRef()`가 실패해도 manifest를 null snapshot fields로 쓰고, 이후 `completed_cut_ids` append 및 `phase3-gate` 전이를 진행합니다 (`hooks/mpl-phase-controller.mjs:772`, `hooks/mpl-phase-controller.mjs:875`). 재현상 git repo가 아닌 workspace에서 `release-finalize`가 `commit_sha:null`, `tree_sha:null`, `snapshot_ref:null`, `artifact_creation_failed:{type:"snapshot_ref"}` manifest를 쓰고 `completed_cut_ids:["mvp"]`, `current_cut_id:null`, `current_phase:"phase3-gate"`로 전진합니다. User-visible artifact 실패는 best-effort로 완료 처리할 수 있지만, snapshot ref는 released-cut immutability의 기준점이라 실패 시 cohort를 완료 처리하면 안 됩니다. snapshot 실패는 manifest write failure처럼 `release-finalize`에 머물고 append를 막는 회귀 테스트가 필요합니다.

확인:
- snapshot 실패 재현: 현재 null snapshot manifest를 쓰고 release 완료 처리됨
- `node --test hooks/__tests__/mpl-release-artifact.test.mjs hooks/__tests__/mpl-phase-controller.test.mjs` 통과 (`96 tests / 0 failures`)
- `npm test` 통과 (`1055 tests / 0 failures`)

— from codex

KyubumShin (commented):
재검토했습니다. 이전에 남긴 snapshot-ref blocker는 해결됐습니다. snapshot ref 생성 실패 시 manifest를 쓰지 않고 `completed_cut_ids` append도 막는 경로와 회귀 테스트를 확인했습니다.

추가로 새 회귀 하나가 있습니다.

1. **High**: `draft_pr` 재시도에서 PR이 실제로 생성되지 않았는데 성공으로 기록될 수 있습니다. `hooks/lib/mpl-release-artifact.mjs:287`의 `branchResult.noop` 분기는 branch가 같은 commit에 있다는 사실만 보고 `gh pr create`를 건너뛰며 `artifact_creation_failed:null`을 반환합니다. 그런데 첫 실행에서 branch 생성/푸시는 성공하고 `gh pr create`만 실패한 경우(gh 미설치, auth 실패, API 실패 등), 다음 실행은 branch noop으로 들어가 `gh`를 다시 호출하지 않고 성공 처리합니다. 결과적으로 manifest의 이전 `draft_pr` 실패가 사라지고 stopReason도 `artifact=draft_pr created (noop)`처럼 보일 수 있지만, 실제 draft PR은 만들어진 적이 없습니다.

재현:
- bare origin이 있는 repo에서 `artifact:'draft_pr'` 실행
- 첫 실행: fake `gh`가 실패하도록 해서 branch push 이후 `gh pr create failed` 기록
- 두 번째 실행: fake `gh`를 성공하도록 바꿔도 현재 코드는 `gh`를 호출하지 않고 `pr_skipped:'branch noop — assumed PR from prior run still open'`로 성공 반환

branch noop은 “release branch가 존재한다”는 증거일 뿐 “draft PR이 이미 존재한다”는 증거가 아닙니다. `draft_pr`는 branch noop이어도 기존 PR을 `gh pr view/list --head <branch>`로 확인하거나, `gh pr create`를 다시 시도하고 “이미 PR 존재” 오류만 idempotent success로 처리하는 쪽이 안전합니다. 회귀 테스트는 “첫 시도에서 branch push 후 `gh pr create` 실패 → 재시도에서 `gh`가 다시 호출되거나 기존 PR 존재가 확인되어야 함”을 보면 잡힙니다.

확인:
- `node --test hooks/__tests__/mpl-release-artifact.test.mjs hooks/__tests__/mpl-phase-controller.test.mjs` 통과 (`105 tests / 0 failures`)
- `npm test` 통과 (`1064 tests / 0 failures`)

— from codex

KyubumShin (commented):
Re-review at `6bf0025`. All five concerns (codex's High + my four) are addressed in this single commit with clean, focused diffs.

**Codex's High** (snapshot bail) was a real RFC §5.4 §232 violation I missed in my own review — I treated snapshot ref as the same "best-effort post-step" as tag/branch/draft_pr, but the RFC explicitly distinguishes them (snapshot is the immutability anchor; user-visible artifacts are external-tool-dependent). The fix at `mpl-phase-controller.mjs:780-789` bails to `release-finalize` with no manifest write and no `completed_cut_ids` append when `createSnapshotRef` returns `ok:false`, and the new test at `mpl-phase-controller.test.mjs:1523-1547` regression-pins the "not a git repo" reproducer. Correct fix; my prior review should have caught this.

Spot-checks on my four:

- **#1 idempotency**: `refExists` helper + same-SHA-noop / different-SHA-refuse split is exactly the contract I asked for. `createArtifactTag` / `createArtifactBranch` both implement it; `attemptArtifactCreation`'s `draft_pr` branch skips `gh pr create` when the branch step is noop (a subtle catch — re-running `gh pr create` against an existing PR would itself fail spuriously). +6 unit tests cover noop same-sha, refuse different-sha, and bare-remote success across all three artifact types, plus the integration test at `mpl-phase-controller.test.mjs:1681-1755` runs release-finalize twice and asserts `doesNotMatch(/tag already exists/i)`. Catches the exact spurious-surface I flagged.

- **#2 single-write order swap**: artifact-attempt now runs at `mpl-phase-controller.mjs:801-816` BEFORE the manifest write at line ~820. The failure result lands in `manifest.artifact_creation_failed` and is captured in the single canonical write — no re-write block, no silent swallow. Cleanest possible resolution; the manifest on disk is now always consistent with the user-visible stopReason.

- **#3 snapshot push**: `createSnapshotRef` now wraps `git update-ref` with a follow-on `tryPushRef`, surfacing `pushed` + `push_reason` on the success path. Push failure is non-fatal (matches the local-ref-is-canonical comment). stopReason tail explicitly shows `(pushed)` vs `(local only, push: <reason>)` so the user knows what made it to the remote. The bare-remote test at `mpl-release-artifact.test.mjs:96-114` proves the push actually lands.

- **#4 HEAD-drift caveat**: docstring at `mpl-release-artifact.mjs:45-52` documents caller responsibility + flags Stage B's likely "explicit invariant" follow-up. Good footnote for the next reader.

### Test bookkeeping

The 1.6c-ii integration tests that previously ran without a git repo now all need `initGitFixture` (since 1.6c-iii's strict snapshot precondition fires before file write). The +5 retrofitted call sites are mechanical and necessary — not a regression. Pre-1.6c-iii placeholder assertions (`manifest.commit_sha === null`) were correctly flipped to `assert.match(.., /^[0-9a-f]{40}$/)`.

1064/1064 hook tests pass.

### Verdict

LGTM at `6bf0025`. The author absorbed both reviews completely and the fix surfaces are scoped to the exact concerns raised — no scope creep, no introduced regressions. Stage A 1.6c complete, ready for merge once codex re-confirms.

— from claude

KyubumShin (commented):
Re-review at `4b634ee`. Codex's round-2 High is a real bug I missed in my first-pass review — and worse, I praised the broken behavior in my prior LGTM as a "subtle catch" that "re-running `gh pr create` against an existing PR would itself fail spuriously." That was wrong: branch-noop only proves the release branch exists, it says nothing about whether `gh pr create` ever ran successfully. The first-run-pushes-branch-but-gh-fails reproducer is concrete and real (gh missing, auth failure, API rate limit are all production paths).

I should have caught this when reviewing the dispatch logic at `attemptArtifactCreation` — the asymmetry between tag/branch (idempotent via `refExists` on the exact same artifact) and `draft_pr` (idempotency derived from a *different* artifact, the branch) was the tell. Apologies for missing it.

### Fix verification

`findExistingDraftPr(cwd, branchName)` at `mpl-release-artifact.mjs:215-235` is exactly the right shape:
- Calls `gh pr list --head <branch> --state open --json url --jq '.[0].url // empty'` — pure observation, never throws, never mutates remote state
- Returns the actual PR URL when one exists, `null` otherwise (including on any gh failure, so the caller falls through to create)

The dispatch update at `attemptArtifactCreation` lines 315-340 unconditionally calls `findExistingDraftPr` after the branch step (no longer gated on `branchResult.noop`):
- PR found → idempotent success with `pr_url` populated in result (better than the prior `pr_skipped` string — downstream tooling can now use the URL)
- PR not found → falls through to `createArtifactDraftPr` regardless of branch noop status

This catches the exact codex reproducer: first run pushes branch + fails at gh pr create → second run sees branch noop but `findExistingDraftPr` returns null → actually calls `createArtifactDraftPr` → PR finally lands.

Test coverage matches the fix:
- Updated existing test now uses a fake `gh` stub that returns a URL for `pr list` and `exit 99` for `pr create` — proves the noop path skips creation only when PR is actually observed.
- New codex-regression test at `mpl-release-artifact.test.mjs:362-415` uses a fake gh that logs every invocation; asserts `gh pr create` IS in the call log after branch noop + empty pr list. Pre-fix would have silently skipped the call and the log assertion would fail.

1065/1065 hook tests pass.

### Verdict

LGTM at `4b634ee`. The fix is correct, the regression test proves the bug stays caught, and the new helper exposes `pr_url` as a bonus side-effect. Stage A 1.6c is genuinely ready to merge once codex re-confirms.

Two-round lesson for me: when one artifact type's idempotency derives from a different artifact's state (PR existence derived from branch presence), that's a red flag to verify the dependency, not paper over it. I'll be sharper on dispatch-layer dependency chains going forward.

— from claude

KyubumShin (commented):
재검토했습니다. 제 이전 High(`draft_pr` 재시도에서 branch noop만 보고 PR 생성 성공으로 오인하는 문제)는 현재 head `4b634ee`에서 해결됐습니다.

확인한 내용:
- `attemptArtifactCreation()`이 `draft_pr` 경로에서 branch noop 여부와 무관하게 `gh pr list --head <branch>`로 기존 open PR을 먼저 확인합니다.
- 기존 PR이 없으면 `gh pr create`를 다시 호출하므로, “첫 실행에서 branch push 성공 후 `gh pr create` 실패 → 두 번째 실행에서 PR 생성 누락” 재현이 더 이상 발생하지 않습니다.
- 제가 남겼던 재현 케이스를 직접 다시 돌렸고, 두 번째 실행에서 `gh pr list`와 `gh pr create`가 모두 호출되며 `artifact_creation_failed:null`은 실제 PR 생성 성공 후에만 반환됩니다.
- 관련 회귀 테스트도 추가되어 있습니다.

추가로 발견한 blocker 없습니다. LGTM.

확인:
- `node --test hooks/__tests__/mpl-release-artifact.test.mjs hooks/__tests__/mpl-phase-controller.test.mjs` 통과 (`106 tests / 0 failures`)
- `npm test` 통과 (`1065 tests / 0 failures`)

— from codex

PR #191, opened 2026-05-25
Add runtime-specific plugin installers

KyubumShin:
## What
- Add runtime-specific installers for Claude Code and Codex CLI.
- Keep Codex marketplace metadata in a generated wrapper that points to this checkout through `./plugins/mpl`.
- Add Codex-compatible `name` fields to MPL skill frontmatter.
- Update install docs and metadata tests for the split installer flow.

## Why
Claude and Codex expect different marketplace layouts. Installing through runtime-specific scripts avoids forcing one shared marketplace structure to satisfy both runtimes and prevents the Codex `./plugins/mpl` convention from conflicting with Claude root-based plugin discovery.

## Design
- `README.md` / `README_ko.md` installation flow
- `docs/design.md` dual-runtime install metadata note
- `docs/standalone.md` runtime manifest table

## Verification
- `HOME=/private/tmp/mpl-installer-smoke-claude ./install/claude.sh`
- `HOME=/private/tmp/mpl-installer-smoke-codex ./install/codex.sh`
- `env HOME=/private/tmp/mpl-installer-smoke-codex codex plugin list`
- `env HOME=/private/tmp/mpl-installer-smoke-codex codex mcp list`
- MCP stdio `listTools()` smoke
- `npm test` (1076 pass / 0 fail)
- `npm run build` in `mcp-server/`
- `claude plugin validate .claude-plugin/plugin.json`
- `claude plugin validate .claude-plugin/marketplace.json`

## Agent Review Status

- [ ] from claude
- [ ] from codex

---


KyubumShin (commented):
First-pass review on the runtime-specific installer split. Overall direction is sound — separating Claude's "register the checkout directly" flow from Codex's "wrap the checkout under `./plugins/mpl`" flow is the right way to stop the two runtimes from fighting over the same marketplace layout. A few concrete issues to address before this is ready to merge.

### Blockers

**1. `agents/mpl-doctor.md` (line 30) is now inconsistent with the new layout.**

The doctor's Category 1 check still says:
> If `.agents/plugins/marketplace.json` exists, verify it exposes plugin `mpl` and the entry points at the **local plugin root**

But after this PR, the repo's `.agents/plugins/marketplace.json` has `source.path: ./plugins/mpl`, which does **not** resolve to the local plugin root (there is no `plugins/` directory at repo root — only inside the wrapper marketplace generated by `install/codex.sh`). Running `/mpl:mpl-doctor` on a healthy install will now misreport this category. `skills/mpl-setup/SKILL.md` was updated in this PR (it now points at `install/codex.sh` and the wrapper convention) but `mpl-doctor.md` was missed. Please update the doctor's check to match the wrapper convention — either accept `./plugins/mpl` as the canonical Codex path, or have the doctor verify `install/codex.sh` instead of the repo-level marketplace.json (mirroring what setup does).

### Non-blocking concerns

**2. Repo-level `.agents/plugins/marketplace.json` is now a non-functional fossil.**

The file ships in the repo with `source.path: ./plugins/mpl`, but the repo has no `plugins/` directory. So if anyone tries the previously-documented `codex plugin marketplace add /path/to/MPL` (pointing Codex directly at the checkout, which the old README explicitly supported), it will silently resolve to a non-existent plugin path. The new README removes that instruction, but the file is still on disk and pinned in place by `hooks/__tests__/mpl-install-metadata.test.mjs` (`entry.source.path === './plugins/mpl'`). Two options:

- Delete `.agents/plugins/marketplace.json` from the repo entirely and let `install/codex.sh` be the single source of truth. The test would then assert the file does **not** exist in the repo (or skip the marketplace.json portion).
- Keep the file but document its role (template, not directly consumable) — e.g., a top-level comment in design.md and a note in the file's containing dir.

Right now it's "exists, is asserted by tests, but cannot be used" — that's the worst of the three states.

**3. The marketplace.json heredoc in `install/codex.sh` (lines 31-52) duplicates the repo file's content.**

If a future change adds a field to `.agents/plugins/marketplace.json` (e.g., `version`, `description`) it has to be manually mirrored in the heredoc. There's no test asserting they stay in sync. Consider either:

- Sourcing the JSON: `cp "${REPO_ROOT}/.agents/plugins/marketplace.json" "${MARKETPLACE_JSON}"` (only safe if option 2 above keeps the file functional).
- Or `jq` it from the repo file with a path-rewrite.
- Or add a test that asserts equality between the two.

**4. Idempotency on re-run is untested.**

`install/codex.sh` runs under `set -euo pipefail`, so a non-zero exit from `codex plugin marketplace add` or `codex plugin add mpl@mpl` will abort. The PR description's smoke tests all used fresh `HOME=/private/tmp/...` directories, so the "already-installed" case wasn't exercised. Same concern for `claude.sh`: `claude plugin marketplace add` against an already-registered marketplace, and `claude plugin install` against an already-installed plugin. If either CLI returns non-zero on duplicate-add, users running the installer twice (e.g., after a `git pull`) will get a confusing failure. Suggest at minimum a comment noting "re-run safe?" expectations, or wrapping the two install calls in a `|| true` with a follow-up `plugin list | grep mpl` verification step.

**5. Minor: `--scope` on `claude.sh`.**

```bash
"${CLAUDE_BIN}" plugin marketplace add --scope "${CLAUDE_SCOPE}" "${REPO_ROOT}"
"${CLAUDE_BIN}" plugin install --scope "${CLAUDE_SCOPE}" mpl
```

Worth a one-line sanity check (`claude plugin marketplace add --help`) to confirm `--scope` is accepted on the marketplace subcommand and not just `plugin install`. The PR description's verification list includes the full smoke install, so this is likely fine, but the test only matches on `plugin marketplace add` (regex `/\$\{CLAUDE_BIN\}" plugin marketplace add/`) — it does **not** assert the `--scope` flag's presence or that the CLI actually accepts it.

### Nits

- `install/codex.sh:23` — error message says "Remove it or set `MPL_CODEX_MARKETPLACE_ROOT` to a different directory." `MPL_CODEX_MARKETPLACE_ROOT` controls the entire marketplace root, not just the plugin symlink — slightly misleading. Fine, but a clearer hint might point to `${PLUGIN_LINK}` specifically.

- The `name:` additions to skill frontmatter (7 files) look correct and the change is minimal. No concern.

— from claude

KyubumShin (commented):
리뷰 코멘트입니다. Claude가 이미 짚은 `agents/mpl-doctor.md` drift, repo-level marketplace 역할, heredoc duplication, idempotency 우려는 중복해서 반복하지 않고, 추가로 확인한 실행 경로 이슈만 남깁니다.

1. **Medium**: `install/codex.sh`가 wrapper의 `plugins/mpl`을 현재 checkout 전체에 symlink한 뒤 `codex plugin add`를 실행해서, Codex가 runtime plugin root를 만들 때 gitignored/local artifacts까지 그대로 cache합니다. 문제 지점은 `install/codex.sh:28-30`의 raw repo symlink입니다. 임시 HOME에서 실제로 `./install/codex.sh`를 실행해 보니 `codex mcp list`의 Cwd가 `/private/tmp/.../.codex/plugins/cache/mpl/mpl/0.18.6/.`로 잡혔고, 그 cache 안에 `.git/`, `mcp-server/node_modules/`, `mcp-server/dist/`, `.mpl/`, `.pr-review-state/`, `.claude/settings.local.json`까지 들어갔습니다. 특히 `tools/mpl-mcp-server-launcher.mjs:29-33`은 `dist/index.js`와 deps marker가 있으면 rebuild를 건너뛰므로, 사용자가 기존 개발 checkout에서 `git pull` 후 installer를 다시 돌리면 ignored/stale `mcp-server/dist`가 새 plugin metadata와 함께 실행될 수 있습니다. 설치 스크립트는 raw checkout을 직접 marketplace plugin root로 노출하지 말고, `git archive`/clean staging directory/명시적 exclude copy 등으로 `.git`, `.mpl`, `.pr-review-state`, `.claude`, `node_modules`, `dist` 같은 local state와 build output을 제외한 plugin root를 만든 뒤 그 경로를 `./plugins/mpl`로 가리켜야 합니다.

확인:
- `env HOME=/private/tmp/mpl-review-codex-smoke-191 CODEX_HOME=/private/tmp/mpl-review-codex-smoke-191/.codex ./install/codex.sh`
- `env HOME=/private/tmp/mpl-review-codex-smoke-191 CODEX_HOME=/private/tmp/mpl-review-codex-smoke-191/.codex codex mcp list`
- cache contents/size check: ignored `mcp-server/node_modules` alone was copied as 123M

— from codex

KyubumShin (commented):
리뷰 피드백 반영했습니다.

Disposition:
- Claude blocker accepted: `agents/mpl-doctor.md` Category 1 now checks `install/codex.sh` wrapper generation instead of repo-root `.agents/plugins/marketplace.json`.
- Claude non-blocking #2/#3 accepted: tracked `.agents/plugins/marketplace.json` was removed, so `install/codex.sh` is the single Codex marketplace source. Design docs and install metadata tests were updated accordingly.
- Claude idempotency concern accepted: both installers now document re-run behavior, and both were verified by running twice against the same fresh `HOME`.
- Codex Medium accepted: `install/codex.sh` no longer exposes the raw checkout as a symlink. It now stages a clean plugin root from `git ls-files` tracked files, with a non-git tar fallback that excludes `.git`, `.mpl`, `.pr-review-state`, `.claude`, `node_modules`, `mcp-server/node_modules`, and `mcp-server/dist`.

Verification:
- `node --test hooks/__tests__/mpl-install-metadata.test.mjs` pass (`4 pass / 0 fail`)
- `npm test` pass (`1076 pass / 0 fail`)
- `npm run build` in `mcp-server/` pass
- `HOME=/private/tmp/mpl-installer-clean-codex ./install/codex.sh` twice pass; `codex plugin list` shows `mpl@mpl installed, enabled`; `codex mcp list` shows `mpl-server enabled`.
- Clean-install artifact check returned no matches for `.git`, `.mpl`, `.pr-review-state`, `.claude`, `mcp-server/node_modules`, or `mcp-server/dist` before MCP first use.
- MCP `listTools()` from the installed clean Codex cache bootstrapped `npm install` + build and returned `mpl_score_ambiguity`, `mpl_state_read`, `mpl_state_write`, `mpl_classify_feature_scope`, `mpl_diagnose_e2e_failure`.
- `HOME=/private/tmp/mpl-installer-clean-claude ./install/claude.sh` twice pass, including metadata validate and already-installed handling.

— from codex

KyubumShin (commented):
Second-pass review against new HEAD `98e07c6` (since first review at `eb6f7c4`).

### First-review items — resolved

- ✅ `agents/mpl-doctor.md` Category 1 updated (commit `51af7f9`). The check now references `install/codex.sh` and the wrapper convention. Good.
- ✅ Repo-level `.agents/plugins/marketplace.json` removed (commit `51af7f9`). The test now asserts `existsSync(...) === false`. `skills/mpl-version-bump/SKILL.md` and `docs/design.md` updated to match. Clean.
- ✅ Heredoc no longer duplicates a repo file. With the file deleted, the installer is the single source of truth.
- ✅ Idempotency comment added to both installers asserting "Codex/Claude CLI accepts already-registered marketplaces". Reasonable, though still untested empirically.

### New blockers introduced by commit `98e07c6` ("Stage clean Codex plugin installs")

The refactor swaps the symlink approach for a **file copy** approach. This is a meaningful behavior change that's worth flagging before merge.

**1. `git pull` no longer auto-updates the Codex view of MPL.**

With the previous symlink, `~/.codex/mpl-marketplace/plugins/mpl` was a symlink to the checkout, so any commit/pull was immediately reflected in Codex's plugin tree. After this PR, that path is a *copy* of the git-tracked files frozen at install time. Users must explicitly re-run `./install/codex.sh` after every `git pull` (or every local edit) to refresh the plugin.

This is a real ergonomic regression for development and for users who track `main` directly. Neither README mentions it, and no installer output hints at it. At minimum, the README should say something like: "After updating MPL (git pull or local edits), re-run `./install/codex.sh` to refresh the staged plugin root."

If the goal of the staging change was to support sandboxed/immutable plugin roots (e.g. Codex doesn't tolerate symlinks well, or wants stable file inodes for caching), please call that out explicitly in the commit message or design.md so future maintainers know why the simpler symlink path was abandoned. Right now the commit message is just "Stage clean Codex plugin installs" with no body — the *why* is invisible.

### Non-blocking concerns

**2. Untracked-but-needed files are silently dropped.**

```bash
while IFS= read -r -d "" REL_PATH; do
  ...
done < <(git -C "${REPO_ROOT}" ls-files -z)
```

`git ls-files` only emits files in the git index. If a developer adds a new skill/hook/etc. and runs `./install/codex.sh` before `git add`-ing it, the file silently won't make it into the staged plugin. They'll get a confusing "skill not found" later. Two options:

- Use `git ls-files -z --others --exclude-standard --cached` to also include untracked-but-not-ignored files (so a fresh `mpl-newskill/SKILL.md` is picked up even before commit).
- Or add a warning when untracked files exist: `git status --porcelain | grep '^??'` → "Warning: untracked files in repo. Run `git add` if you want them staged."

The current behavior is technically correct (only ship what's committed) but the silence is surprising during development.

**3. Tar fallback is a deny-list of files to exclude.**

```bash
tar -C "${REPO_ROOT}" \
  --exclude "./.git" --exclude "./.mpl" --exclude "./.pr-review-state" \
  --exclude "./.claude" --exclude "./node_modules" \
  --exclude "./mcp-server/node_modules" --exclude "./mcp-server/dist" \
  --exclude "./.DS_Store" \
  -cf - . | tar -x -C "${PLUGIN_TMP}"
```

This branch only runs when `REPO_ROOT` is not a git work tree (rare — but possible if someone unpacks a release tarball). When it does run, it's a "everything except this enumerated list" policy. Real-world scratch artifacts that wouldn't be in the list:

- `coverage/`, `.nyc_output/`, `.cache/`, IDE state (`.vscode/`, `.idea/`)
- Any user-dropped scratch files at the repo root
- The PLUGIN_TMP itself if `MPL_CODEX_MARKETPLACE_ROOT` is set inside REPO_ROOT (unlikely, but a self-recursion footgun)

A safer approach: ship a `.gitattributes` file with `export-ignore` markers and use `git archive` even from the tarball path (most tarballs are produced by `git archive` and contain a manifest). Or replace the tar branch with a hard error: "MPL must be installed from a git work tree. Found a non-git directory: $REPO_ROOT" — this is fine since the README says `git clone`, so non-git is an unexpected state.

**4. Small destructive window: `rm -rf "${PLUGIN_ROOT}"` then `mv "${PLUGIN_TMP}" "${PLUGIN_ROOT}"`.**

```bash
rm -rf "${PLUGIN_ROOT}"
mv "${PLUGIN_TMP}" "${PLUGIN_ROOT}"
```

If the `mv` fails for any reason (disk full, race, ENOENT after a sibling process touched the path), the user is left with no plugin installed. The previous symlink approach was atomic (`ln -s` overwrites). Suggest swap-then-delete:

```bash
if [ -e "${PLUGIN_ROOT}" ]; then
  mv "${PLUGIN_ROOT}" "${PLUGIN_ROOT}.old.$$"
fi
mv "${PLUGIN_TMP}" "${PLUGIN_ROOT}"
rm -rf "${PLUGIN_ROOT}.old.$$"
```

Small change, eliminates the destructive window.

**5. Test is static regex over the shell script text.**

```js
assert.match(codexInstaller, /git -C .*\$\{REPO_ROOT\}.*ls-files -z/);
assert.match(codexInstaller, /cp -p .*\$\{REPO_ROOT\}\/\$\{REL_PATH\}/);
assert.doesNotMatch(codexInstaller, /ln -s/);
```

These assertions tell you the shell script *contains the right tokens*, not that the script *does the right thing*. A genuine smoke test would set `MPL_CODEX_MARKETPLACE_ROOT=$(mktemp -d)` + `CODEX_BIN=$(mktemp ... && chmod +x)` (stub binary) and assert the staged plugin tree contains expected files. Not blocking, but the regex tests are fragile — they'll break on any innocuous refactor (e.g., renaming `REL_PATH` to `REL` or moving `--exclude` to a variable) even if behavior is unchanged.

**6. MCP server `dist/` is dropped from staging.**

This is intentionally excluded (it's a build artifact). But after every re-install, the launcher (`tools/mpl-mcp-server-launcher.mjs`) needs to rebuild on first MCP invocation. That means after `git pull && ./install/codex.sh`, the *first* Codex session that touches the MCP server will pay a `npm install && npm run build` delay (and the spinner may not be obvious). Worth mentioning in the install completion message:

```bash
echo "[MPL] Codex install complete. Start a new Codex session to load the MPL plugin."
echo "[MPL] Note: the MCP server will rebuild on first use (one-time, ~30s)."
```

**7. `MARKETPLACE_JSON` heredoc could still drift** — from the schema Codex actually expects. With the repo file deleted, there's no schema fixture to validate against. If Codex's marketplace schema changes (new required field, renamed key), only the installer needs to update — but there's no test that catches the divergence. Consider versioning the heredoc with a comment like `# Codex marketplace schema v1 (matches docs/design.md §Install Metadata)`.

### Nits

- `install/codex.sh:12` — `PLUGIN_TMP="${MARKETPLACE_ROOT}/plugins/.mpl.tmp.$$"`. The `$$` (PID) guards against collision on sequential runs but not concurrent ones (two terminals running the installer simultaneously). Two concurrent runs would have different PIDs and stage into different temp dirs, but both would then `mv` into `${PLUGIN_ROOT}` — the second wins. Edge case, but worth a `flock` or refusing concurrent runs.
- The two installers don't share any helpers — `command -v "${BIN}" || error` and the validation echo pattern are duplicated. Probably not worth extracting yet (only 2 installers, 30 lines each) but worth keeping in mind if a third runtime appears.

— from claude

KyubumShin (commented):
Claude second-pass 리뷰 반영했습니다.

Disposition:
- Blocker accepted: README/README_ko, installer output, design doc에 Codex staged root는 `git pull` 또는 local edits 이후 `./install/codex.sh` 재실행이 필요하다고 명시했습니다.
- Rationale accepted: design doc에 symlink 대신 clean tracked-file staging을 쓰는 이유를 추가했습니다. local state/build artifacts가 Codex plugin cache로 들어가지 않게 하려는 의도입니다.
- Non-blocking #2 accepted: `install/codex.sh`가 untracked-but-not-ignored files를 감지하면 staged root에 포함되지 않는다고 경고합니다.
- Non-blocking #3 accepted: non-git tar fallback은 제거하고, Codex install은 MPL git checkout에서만 동작하도록 fail-fast 처리했습니다.
- Non-blocking #4 accepted: staged root 교체를 `old -> new -> cleanup` 방식으로 바꿔 `rm -rf root` 후 `mv`의 destructive window를 줄였습니다.
- Non-blocking #5 accepted: install metadata test에 stub Codex CLI를 이용한 실제 staging smoke test를 추가했습니다.
- Non-blocking #6/#7 accepted: installer completion message에 MCP first-use prepare/build를 알리고, heredoc 앞에 Codex marketplace schema comment를 추가했습니다.
- Nit partially accepted: simple lock directory로 concurrent installer 실행을 차단합니다.

Verification:
- `node --test hooks/__tests__/mpl-install-metadata.test.mjs` pass (`5 pass / 0 fail`)
- `npm test` pass (`1077 pass / 0 fail`)
- `npm run build` in `mcp-server/` pass
- `HOME=/private/tmp/mpl-installer-review4-codex ./install/codex.sh` twice pass; `codex plugin list` shows `mpl@mpl installed, enabled`; `codex mcp list` shows `mpl-server enabled`.
- Install artifact check returned no install-time matches for `.git`, `.mpl`, `.pr-review-state`, `.claude`, `mcp-server/node_modules`, or `mcp-server/dist`.
- MCP `listTools()` from clean Codex cache bootstrapped first-use install/build and returned all 5 MPL tools.
- `HOME=/private/tmp/mpl-installer-review4-claude ./install/claude.sh` twice pass.
- `git diff --check` pass.

— from codex

KyubumShin (commented):
README 갱신 안내를 추가로 정리했습니다.

- `README.md`: Codex staged root refresh note를 별도 블록으로 분리했습니다.
- `README_ko.md`: 빠른 시작과 설치 섹션 모두에 `git pull`/local edits 이후 `./install/codex.sh` 재실행 필요성을 명시했습니다.

Verification:
- `node --test hooks/__tests__/mpl-install-metadata.test.mjs` pass (`5 pass / 0 fail`)
- `git diff --check` pass

— from codex

PR #192, opened 2026-05-25
Add gitless curl bootstrap installer

KyubumShin:
## What
- Add a top-level `install.sh` bootstrap that supports `--runtime claude|codex|both`.
- Support gitless installs by downloading a GitHub source archive with `curl`, persisting it under `~/.mpl/install/source/mpl`, and writing `.mpl-install-manifest`.
- Let Codex stage from the manifest when no git checkout exists, while keeping git checkout staging unchanged.
- Move Claude MCP config to the shared launcher so archive installs can build `mcp-server` on first use.
- Update README/README_ko, standalone/design docs, setup/doctor guidance, and installer metadata tests.

## Why
Claude Code and Codex CLI should both be installable with a downloaded `install.sh` flow, without requiring users to have git or forcing both runtimes into one marketplace format.

## Design
- `docs/design.md`
- `docs/standalone.md`

## Verification
- `bash -n install.sh`
- `bash -n install/codex.sh`
- `node --test hooks/__tests__/mpl-install-metadata.test.mjs` (6 pass)
- `npm test` (1078 pass)
- Real archive install smoke with `./install.sh --runtime both` using `MPL_TARBALL_PATH`: Claude + Codex install completed; `codex plugin list` shows `mpl@mpl installed, enabled`; `codex mcp list` shows `mpl-server enabled`.
- MCP first-use smoke from the installed Codex plugin cache returned `mpl_score_ambiguity`, `mpl_state_read`, `mpl_state_write`, `mpl_classify_feature_scope`, `mpl_diagnose_e2e_failure`.
- Source/staged roots checked clean of `.git`, `.mpl`, `.pr-review-state`, `.claude`, `mcp-server/node_modules`, and `mcp-server/dist` before MCP first use.
- `git diff --check`

## Agent Review Status

- [ ] from claude
- [ ] from codex

---


KyubumShin:
curl 기반 gitless 설치 흐름을 새 PR로 분리해 반영했습니다.

확인한 내용:
- `install.sh`가 `--runtime claude|codex|both`를 지원하고, checkout이 아니면 GitHub archive를 내려받아 persistent source root에 배치합니다.
- archive/local non-git source에는 `.mpl-install-manifest`를 생성하고, Codex installer는 git이 없어도 manifest 기준으로 clean staged root를 만듭니다.
- Claude MCP 설정도 shared launcher로 바꿔 archive install에서 prebuilt `mcp-server/dist`가 없어도 첫 MCP 호출 시 build됩니다.
- README/README_ko, standalone/design docs, setup/doctor guidance, install metadata tests를 curl bootstrap 기준으로 갱신했습니다.

검증:
- `bash -n install.sh`
- `bash -n install/codex.sh`
- `node --test hooks/__tests__/mpl-install-metadata.test.mjs` (6 pass)
- `npm test` (1078 pass)
- 실제 archive smoke: `./install.sh --runtime both`로 Claude + Codex 설치 완료, `codex plugin list`에서 `mpl@mpl installed, enabled`, `codex mcp list`에서 `mpl-server enabled`
- 설치된 Codex plugin cache에서 MCP first-use 호출 성공: `mpl_score_ambiguity`, `mpl_state_read`, `mpl_state_write`, `mpl_classify_feature_scope`, `mpl_diagnose_e2e_failure`
- source/staged root에는 `.git`, `.mpl`, `.pr-review-state`, `.claude`, `mcp-server/node_modules`, `mcp-server/dist`가 섞이지 않는 것 확인
- `git diff --check`

— from codex

KyubumShin (commented):
First-pass review on the gitless curl bootstrap (PR #192, HEAD `ca67847`). Direction is solid — the persistent `~/.mpl/install/source/mpl` root + manifest-based staging is a clean answer to the symlink-vs-copy tradeoff from #191, and the MCP launcher swap makes archive installs work without committed build output. A few real bugs and UX concerns to address before merge.

### Blockers

**1. `write_manifest` self-reference bug for `SOURCE_KIND="local"` re-runs (`install.sh:105-114`).**

```bash
write_manifest() {
  local source_dir="$1"
  ...
  ( cd "${source_dir}"; find . -type f -print | LC_ALL=C sort | sed "s#^\./##" ) >"${manifest_tmp}"
  mv "${manifest_tmp}" "${source_dir}/.mpl-install-manifest"
}
```

For the archive path this is correct — `download_source` extracts to a fresh `extract_dir` with no preexisting `.mpl-install-manifest`, so `find` doesn't see it. But for `SOURCE_KIND="local"` (extracted tarball without git, line 170-171):

- First run: `.mpl-install-manifest` doesn't exist in `SOURCE_ROOT` → `find` doesn't include it → clean manifest. ✓
- **Second run**: `.mpl-install-manifest` exists from the first run → `find` includes it → manifest references itself → `install/codex.sh`'s manifest reader copies `.mpl-install-manifest` into the staged Codex plugin root.

The test (`hooks/__tests__/mpl-install-metadata.test.mjs:316`) asserts `existsSync(stagedRoot/.mpl-install-manifest) === false`, but only on a first-run archive scenario — it doesn't exercise the local-non-git re-run case, so CI won't catch this.

Fix: either exclude the manifest from `find` (`find . -type f ! -name '.mpl-install-manifest'`) or remove it first (`rm -f "${source_dir}/.mpl-install-manifest"`) before regenerating.

**2. `--ref` is silently ignored when running from a local checkout (`install.sh:165-176`).**

```bash
if [ "${MPL_FORCE_DOWNLOAD}" != "1" ] && [ -n "${SCRIPT_DIR}" ] && [ -f "${SCRIPT_DIR}/.claude-plugin/plugin.json" ] ... ; then
  SOURCE_ROOT="${SCRIPT_DIR}"
  ...
else
  download_source
fi
```

A user who runs `./install.sh --ref v0.18.3` from a git checkout will get the *checkout's* state, not the ref'd state. The `--ref` argument is parsed and stored in `MPL_REF` but never used in the local-source branch. This is a real surprise: someone trying to pin a stable version from a clone gets the latest local commits instead.

Three reasonable options:
- Warn at the top of the local-source branch: `if [ "${MPL_REF}" != "main" ]; then echo "[MPL] Warning: --ref ${MPL_REF} ignored — using local checkout. Set MPL_FORCE_DOWNLOAD=1 to download instead." >&2; fi`
- Force download when `--ref` was explicitly passed (track it via a separate flag, e.g. `MPL_REF_EXPLICIT=1`).
- Document the precedence in `usage()` and the README.

Of these I'd pick the warning — least surprising, minimal code, lets advanced users override with `MPL_FORCE_DOWNLOAD=1`.

### Non-blocking concerns

**3. `MPL_REF` defaults to `main` (unpinned).**

```bash
MPL_REF="${MPL_REF:-main}"
```

Every fresh curl-bootstrap install downloads whatever is on `main` at that moment. This is fine for early-adopter / continuous-deployment use, but for users following the README's curl one-liner it means:

- No reproducibility — two users installing on different days get different code.
- Any malicious commit landing on `main` (compromised maintainer credential, etc.) flows immediately into all new installs.
- Bisecting an install-time bug is painful.

Industry pattern (rustup, brew, etc.) is to default the install script to the latest *tagged* release, not HEAD. Suggest:

- Add a build step that updates `MPL_REF` default to the latest release tag (or hardcode the latest version at version-bump time).
- Or document loudly in the README: "For production use, pin to a release tag: `... | bash -s -- --ref v0.18.4`."

**4. Security: curl-piped install has no integrity verification.**

The README's first-class install command is `curl ... install.sh | bash`. This is standard OSS practice but has known downsides:

- No SHA256 / GPG check on `install.sh` itself.
- No checksum on the downloaded source tarball (curl uses HTTPS so transit is safe, but a compromised GitHub CDN edge or a registrar-level attack could serve poisoned bytes).
- `MPL_REF=main` (concern #3 above) means even the legitimate-content attack surface is "whoever has push to main".

Mitigations worth adding (not blockers):

- A README note: "To verify before running, download install.sh, inspect, then execute: `curl -o install.sh ... && less install.sh && bash install.sh ...`."
- Optional: support `MPL_TARBALL_SHA256=<hash>` env var that, if set, is checked against the downloaded archive before `tar -xzf`.

**5. MCP config swap (`.mcp.json`) is a quiet upgrade boundary.**

```diff
-      "args": ["${CLAUDE_PLUGIN_ROOT}/mcp-server/dist/index.js"]
+      "args": ["${CLAUDE_PLUGIN_ROOT}/tools/mpl-mcp-server-launcher.mjs"]
```

Existing Claude users with prebuilt `mcp-server/dist/` will continue to work *until they re-run install.sh*, at which point:

- The new `.mcp.json` is registered via the marketplace.
- The launcher runs on next MCP call — even though `dist/` exists locally, the launcher may rebuild (depending on its freshness check).

Worth checking that the launcher's "skip rebuild if dist is fresh" logic actually works for upgraders. If it always rebuilds, every user upgrading via `install.sh` will pay a one-time `npm install && npm run build` delay on the next MCP call. Not a bug, but worth a sentence in the README's upgrade flow: "Existing installs will pick up the new MCP launcher after re-running `install.sh`. The first MCP call after upgrade may rebuild the MCP server."

**6. `usage()` is out of sync with the implementation.**

- `--repo` is parsed (line 50-54) but absent from the usage string.
- `MPL_FORCE_DOWNLOAD` is checked at line 165 but not listed in the Environment section. If this is test-only, mark it `(internal — do not rely on)` in a comment near where it's used, or remove it from the public surface entirely.
- `MPL_TARBALL_URL` and `MPL_TARBALL_PATH` are env-only — consider matching `--tarball-url` / `--tarball-path` flags for consistency, or document why they're env-only.

**7. Auto-mode silently installs both when both CLIs are present.**

```bash
auto)
  if command -v "${CLAUDE_BIN:-claude}" >/dev/null 2>&1; then install_claude=1; fi
  if command -v "${CODEX_BIN:-codex}" >/dev/null 2>&1; then install_codex=1; fi
```

A user who only wants Codex but happens to have `claude` in PATH (e.g., a previous Claude Code install they don't actively use) will get both installed. Worth a one-line `echo "[MPL] Auto-detected: installing for claude, codex. Use --runtime to override."` before the install loop.

**8. `download_source` cleanup function is leaked into global scope.**

```bash
download_source() {
  ...
  cleanup_download() {
    rm -rf "${tmp_dir}"
  }
  trap cleanup_download EXIT
  ...
}
```

Bash function definitions are always global, even when textually nested. After `download_source` returns, `cleanup_download` is still in the shell's function table. Harmless but worth either making the cleanup inline (`trap "rm -rf '${tmp_dir}'" EXIT`) or naming it `__mpl_cleanup_download` to signal "internal".

### Nits

- `install.sh:9` — `MPL_FORCE_DOWNLOAD` is referenced once at line 165. If only the test uses it, consider renaming to `_MPL_FORCE_DOWNLOAD` (leading underscore) or moving to a dedicated test-fixture file. Right now it looks like part of the public API.

- `install.sh:175` (`download_source`) doesn't return a status; the caller assumes success because of `set -e`. Fine, but readers tracing the SOURCE_ROOT assignment have to scan two functions to confirm it's set. A `return 0` at the end of `download_source` (after `trap - EXIT`) would make the intent slightly clearer.

- No concurrent-run protection on `install.sh` itself. The codex installer has `LOCK_DIR` but the orchestrator doesn't. Two simultaneous `install.sh` runs could race on `replace_dir`. Edge case, not worth fixing.

- The `--repo` flag accepts arbitrary input that's interpolated into a URL (`https://github.com/${MPL_REPO}/archive/...`). If a malicious caller invokes `./install.sh --repo "evil/repo"`, install.sh happily fetches from that repo. This is correct behavior (it's a CLI flag the caller chose), but worth a one-line guard against accidental shell metacharacters: `case "${MPL_REPO}" in *[^A-Za-z0-9_./-]*) echo "error: invalid --repo" >&2; exit 1 ;; esac`.

— from claude

KyubumShin:
Claude review 대응 완료했습니다. 새 head: `7e79756`.

Disposition:
- **Blocker 1 accepted**: `write_manifest`가 `.mpl-install-manifest` 자체를 제외하도록 수정했습니다. local non-git source에서 `install.sh --runtime codex`를 두 번 실행해도 staged Codex root에 manifest가 복사되지 않는 회귀 테스트를 추가했습니다.
- **Blocker 2 accepted**: `--ref`/`MPL_REF`가 명시된 상태로 local source를 쓰면 경고를 출력합니다. 특정 ref 다운로드가 필요하면 `MPL_FORCE_DOWNLOAD=1`을 사용하도록 안내합니다.
- **Non-blocking 3 accepted as docs**: README/README_ko에 reproducible install은 release tag를 pin하라고 명시했습니다. 기본값을 latest tag로 바꾸는 release automation은 별도 정책 변경이라 이번 PR에서는 문서화로 처리했습니다.
- **Non-blocking 4 accepted**: `MPL_TARBALL_SHA256` 검증을 추가했고, README에 inspect-before-run 흐름을 추가했습니다.
- **Non-blocking 5 accepted**: 기존 설치가 rerun 후 shared MCP launcher를 사용하며 첫 MCP 호출에서 rebuild될 수 있음을 README/README_ko에 명시했습니다.
- **Non-blocking 6 accepted**: `usage()`에 `--repo`, `MPL_FORCE_DOWNLOAD`, `MPL_TARBALL_SHA256`를 반영했습니다. `MPL_TARBALL_URL`/`MPL_TARBALL_PATH`는 advanced/offline env-only surface로 유지했습니다.
- **Non-blocking 7 accepted**: auto runtime detection 결과를 명시적으로 echo합니다.
- **Non-blocking 8 accepted**: cleanup function을 internal name으로 바꾸고 정상 종료 시 unset합니다.
- **Nits**: `MPL_FORCE_DOWNLOAD`는 public override로 usage에 문서화, `download_source`에 `return 0` 추가, `--repo` 입력 guard 추가. top-level install lock은 reviewer도 edge case로 본 항목이라 이번 PR에서는 deferred로 둡니다.

검증:
- `bash -n install.sh`
- `bash -n install/codex.sh`
- `node --test hooks/__tests__/mpl-install-metadata.test.mjs` (8 pass)
- `npm test` (1080 pass)
- 실제 archive smoke: `MPL_TARBALL_SHA256=<sha> ./install.sh --runtime both` 경로로 Claude + Codex 설치 완료, Codex plugin installed/enabled, MCP enabled 확인
- 설치된 Codex plugin cache에서 MCP first-use `listTools` 성공
- staged Codex root에 `.git`, `.mpl`, `.pr-review-state`, `.claude`, `.mpl-install-manifest`, `mcp-server/node_modules`, `mcp-server/dist`가 섞이지 않는 것 확인
- `git diff --check`

GitHub checks: 현재 branch에 reported checks 없음.

— from codex

KyubumShin (commented):
Second-pass review on PR #192 against HEAD `7e79756` (since first review at `ca67847`).

### Prior items — all addressed

- ✅ **Blocker 1** (`write_manifest` self-reference): `find` now excludes `.mpl-install-manifest`, and the new test `keeps local gitless reruns from staging the bootstrap manifest` (`mpl-install-metadata.test.mjs:221-263`) is a genuine behavioral regression test — it tars a source, extracts to a non-git local dir, runs `install.sh --runtime codex` twice, and asserts neither the manifest content nor the staged Codex root contains the self-reference. That's exactly the right coverage. Solid fix.
- ✅ **Blocker 2** (`--ref` silently ignored): `MPL_REF_PROVIDED` is tracked from both env (`${MPL_REF+x}` test) and flag, warning printed when running from local source, README points users at `MPL_FORCE_DOWNLOAD=1` for the override path. Clear UX.
- ✅ **Concern 3** (unpinned `MPL_REF=main`): documentation in README/README_ko now explicitly encourages release tag pinning with a concrete example. Defaulting to the latest release tag was deferred to a separate policy change; reasonable to scope out.
- ✅ **Concern 4** (curl-pipe-bash integrity): `MPL_TARBALL_SHA256` env var, fallback between `sha256sum` and `shasum`, checked before tar extraction. README now documents the inspect-before-run flow. Good.
- ✅ **Concern 5** (MCP launcher upgrade boundary): README/README_ko refresh notes call out that existing installs pick up the shared launcher after rerun and may rebuild on first MCP call.
- ✅ **Concern 6** (`usage()` doc gaps): `--repo`, `MPL_FORCE_DOWNLOAD`, `MPL_TARBALL_SHA256` all in `usage()`. `MPL_TARBALL_URL`/`MPL_TARBALL_PATH` retained as env-only — fine given the author's rationale (advanced/offline use).
- ✅ **Concern 7** (auto-mode silent dual install): three explicit echo paths now distinguish "both detected", "claude only", "codex only" with a hint about `--runtime`.
- ✅ **Concern 8** (cleanup function global leak): renamed to `__mpl_cleanup_download`, explicit `unset -f` after the trap is cleared. Clean.
- ✅ **Nits**: `MPL_FORCE_DOWNLOAD` documented as a public override (sensible, given the warning path advises it). `download_source` ends in `return 0`. `--repo` input is validated against a tight character whitelist + structural pattern check. The concurrent-install lock was reasonably deferred.

I traced the `MPL_REPO` validation patterns by hand — `*/*/*|""|/*|*/|*//*|*[!A-Za-z0-9_./-]*` correctly rejects multi-component paths, empty, leading/trailing slash, double slash, and out-of-whitelist characters, with `*/*` catching the valid owner/repo shape and the trailing `*)` catching no-slash inputs. Tight.

### Minor follow-ups (non-blocking)

**1. `MPL_REF_PROVIDED` triggers the warning even when `MPL_REF` is explicitly set to the default `main`.**

```bash
MPL_REF_PROVIDED=0
if [ "${MPL_REF+x}" = "x" ] && [ -n "${MPL_REF}" ]; then
  MPL_REF_PROVIDED=1
fi
```

A user who passes `MPL_REF=main ./install.sh` from a local checkout (perhaps for documentation/CI clarity) will see the "ignored" warning even though they didn't ask for anything different from the default. Low-noise edge case, but easy to silence:

```bash
if [ "${MPL_REF_PROVIDED}" = "1" ] && [ "${MPL_REF}" != "main" ]; then
  echo "[MPL] Warning: ..." >&2
fi
```

Or compare against the resolved branch if it's a git checkout. Not worth blocking on, just a UX wart.

**2. No behavioral test for `MPL_TARBALL_SHA256` verification.**

The `documents bootstrap installer safeguards` test (`mpl-install-metadata.test.mjs:117-127`) only asserts the string `MPL_TARBALL_SHA256` appears in the script. Worth adding two behavioral cases:

- Positive: compute the SHA of the tarball generated by `makeSourceArchive`, set `MPL_TARBALL_SHA256` to it, run install.sh, assert success.
- Negative: set `MPL_TARBALL_SHA256` to a wrong value, expect non-zero exit and the "archive SHA256 mismatch" message in stderr.

This is the security-critical path; static string presence doesn't catch e.g., a typo where the script reads from the wrong variable or skips the check when the env is set.

**3. The `MPL_REPO` validation case is correct but dense (`install.sh:84-95`).**

```bash
case "${MPL_REPO}" in
  */*/*|""|/*|*/|*//*|*[!A-Za-z0-9_./-]*)
    ...
```

A one-line comment above the case would help future readers ("Reject empty, leading/trailing slash, more than one slash, double slash, or non-`[A-Za-z0-9_./-]` chars; accept exactly `owner/repo`."). Inscrutable patterns are a perennial source of "what does this actually reject?" questions.

**4. Archive install always re-downloads on rerun.**

`download_source` unconditionally fetches the tarball, even when `MPL_SOURCE_DIR` already exists with the same `MPL_REF`. For users on slow networks updating multiple times a day, this is wasted bandwidth. A simple guard:

```bash
if [ -f "${MPL_SOURCE_DIR}/.mpl-install-manifest.ref" ] \
   && [ "$(cat "${MPL_SOURCE_DIR}/.mpl-install-manifest.ref")" = "${MPL_REF}" ] \
   && [ -z "${MPL_TARBALL_PATH}" ] \
   && [ -z "${MPL_TARBALL_URL}" ]; then
  echo "[MPL] Source at ${MPL_SOURCE_DIR} already matches ref ${MPL_REF}, skipping download. Use MPL_FORCE_DOWNLOAD=1 to refetch."
  SOURCE_ROOT="${MPL_SOURCE_DIR}"
  SOURCE_KIND="archive"
else
  download_source  # writes the .ref file at the end
fi
```

But this also means stale `main` could persist if `main` advances. Not a slam-dunk improvement — feel free to defer. The current "always refetch" behavior is correct, just potentially wasteful.

### Disposition

Blockers resolved cleanly with proper regression coverage. The remaining items above are minor improvements, not gates. From the claude side this looks ready to merge once the gate is satisfied.

— from claude

PR #193, opened 2026-05-25
Add interactive Claude install scope selection

KyubumShin:
## What
- Add Claude install scope parsing to the top-level bootstrap and direct Claude installer: `user`, `project`, `local`, and `ask`.
- Implement `--scope ask` as a Bash prompt that reads from `/dev/tty`, so `curl | bash` still works and Enter keeps the default `user` scope.
- Pass selected Claude scope through gitless/archive installs and update README, Korean README, standalone docs, and setup skill guidance.
- Extend install metadata tests for the interactive prompt contract and explicit scope pass-through.

## Why
Claude Code defaults plugin installs to user scope, but MPL installs need an explicit way to choose user, project, or local scope during gitless curl installs. The interactive `ask` mode keeps the one-line installer ergonomic without requiring git.

## Design
- `docs/standalone.md`
- `README.md`
- `README_ko.md`

## Verification
- `bash -n install.sh`
- `bash -n install/claude.sh`
- `node --test hooks/__tests__/mpl-install-metadata.test.mjs`
- `npm test`
- `git diff --check`

## Agent Review Status

- [x] from claude
- [x] from codex

---


KyubumShin (commented):
PR #193 로컬 Codex 리뷰입니다. 변경 범위는 Claude 설치 scope 선택과 문서/테스트 업데이트에 한정되어 있고, blocker로 볼 문제는 발견하지 못했습니다.

확인한 내용:
- `install.sh`와 `install/claude.sh` 모두 `--scope user|project|local|ask`를 validation하고, `ask`는 `/dev/tty`에서 읽도록 되어 있어 `curl | bash`의 stdin 충돌을 피합니다.
- non-interactive 환경에서 `--scope ask`는 명시 scope를 쓰라는 에러로 종료하므로 자동화 기본값은 유지됩니다.
- 선택된 Claude scope가 direct installer와 gitless archive bootstrap 경로 모두에 전달되는 테스트가 추가되어 있습니다.

리스크:
- 실제 interactive prompt 입력 자체는 자동 테스트에서 직접 누르지 않고 static contract로 고정되어 있습니다. 이건 터미널 대기 리스크를 피하기 위한 합리적인 선택으로 봅니다.

테스트: 작성자가 보고한 `npm test`, `bash -n`, `git diff --check` 결과를 PR 본문과 로컬 상태 기준으로 확인했습니다. 추가 blocker 없음. LGTM.

— from codex

KyubumShin (commented):
Here are my findings for PR #193:

---

FINDING 1 - Code Duplication / Maintainability Risk
install.sh and install/claude.sh both implement identical prompt_claude_scope() functions. Any future change (new scope value, different prompt wording) must be kept in sync manually. The existing pattern already delegates Claude installation to install/claude.sh, so install.sh could simply export MPL_CLAUDE_SCOPE=ask and let claude.sh own the prompt logic, or extract the function into a shared lib file sourced by both. As-is, a future patch that adds a fourth scope will have a silent mismatch until the tests catch it.

FINDING 2 - Shell Portability: local keyword in POSIX sh context
install.sh declares `#!/bin/sh` (inferred from the set -eu preamble and POSIX-style conditionals throughout the existing file). The `local` keyword is not defined by POSIX sh. It works on bash, dash (as an extension), and most modern shells, but it is not portable to strict POSIX sh. Both `prompt_claude_scope` functions use `local choice=""`. Check the shebang of install.sh; if it is `#!/bin/sh`, replace `local choice=""` with `choice=""` to be safe, or change the shebang to `#!/usr/bin/env bash` explicitly.

Files: install.sh line ~122, install/claude.sh line ~52

FINDING 3 - --scope ask over curl|bash: /dev/tty check is correct but docs are inconsistent
The docs and README examples promote `--scope ask` as the recommended invocation in the curl|bash snippets:

  README.md: "curl ... | bash -s -- --runtime claude --scope ask"
  README_ko.md: same

But a curl|bash pipe makes stdin the network socket, not the terminal. The code correctly falls back to /dev/tty, and the check `[ ! -r /dev/tty ] || [ ! -w /dev/tty ]` guards against headless environments. However, the docs say "Enter keeps user scope" as if it is the normal expected flow for the piped command, which may confuse users who copy-paste into a CI script and hit the "requires an interactive terminal" error. Consider keeping the simpler `--scope user` as the default documented example and showing `--scope ask` as a secondary option, or add a note that `--scope ask` requires a real TTY even in the piped form.

FINDING 4 - Test: claude stub test does not verify HOME isolation
The new test "passes selected Claude scope through the direct Claude installer" sets HOME=tempRoot, which is correct for isolation. However, it does not assert that no files were written outside tempRoot (e.g., to the real ~/.mpl or to the real Claude config). If the installer were to accidentally ignore CLAUDE_BIN and fall through to a real `claude` binary, the test would still pass (or fail with an unrelated error) rather than providing a clear signal. Consider adding an assertion that no unexpected side effects occurred, or add a negative assertion that the real ~/.mpl directory was not modified.

hooks/__tests__/mpl-install-metadata.test.mjs line ~155

FINDING 5 - install/claude.sh: CLAUDE_SCOPE initialized from MPL_CLAUDE_SCOPE before arg parsing
install/claude.sh line 8: `CLAUDE_SCOPE="${MPL_CLAUDE_SCOPE:-user}"` runs before the arg-parsing loop. If someone passes `--scope local` on the CLI, the loop correctly overwrites CLAUDE_SCOPE. But the validation case-statement runs after the loop, so it correctly sees the final value. This ordering is fine functionally. However, it creates a subtle precedence where the env var wins silently if the user mistakenly exports MPL_CLAUDE_SCOPE and also passes --scope; actually the CLI arg wins because the loop assigns last. Document this precedence explicitly in the usage() output or in a comment so future maintainers do not reverse the order.

install/claude.sh lines 8-32

FINDING 6 - Missing test: --scope ask with non-TTY in install.sh path
There is a test for the direct claude.sh installer with an explicit scope, but no test exercises the `--scope ask` code path through install.sh when /dev/tty is unavailable (should print the error and exit 1). The existing metadata test suite checks for string presence but does not invoke the function. A small execFileSync test with a closed TTY would close this gap.

hooks/__tests__/mpl-install-metadata.test.mjs

---

None of the above are data-loss or security issues. Findings 1 and 2 are the most actionable before merge. Findings 3-6 are quality/hardening items.

— from claude

KyubumShin (commented):
Here is the review of PR #193 at head f52eb0abdb8b89b629f3ebe3b70c6a79415522ec.

---

**1. install/claude.sh — scope validation fires too late when MPL_CLAUDE_SCOPE is "ask" but TTY is absent**

In install/claude.sh the flow is:
1. Parse args and accept "ask" past the case validator (that is correct).
2. Call prompt_claude_scope only if CLAUDE_SCOPE == "ask".
3. Inside prompt_claude_scope, check /dev/tty readability and exit 1 if not available.

The problem is subtle: prompt_claude_scope exits from a subshell context — it's called as a plain function, not a subshell, so exit 1 does bubble up. That part is fine. However, the /dev/tty guard uses:

  if [ ! -r /dev/tty ] || [ ! -w /dev/tty ]; then

The -r/-w tests on /dev/tty can return false in some CI environments even when the FD is technically "open" (e.g. Docker with /dev/tty allocated but not a real tty), causing prompt_claude_scope to fall through to the read -r loop and hang indefinitely. The more robust check is:

  if ! tty -s < /dev/tty 2>/dev/null; then

or equivalently check that /dev/tty can actually be opened as a terminal:

  if ! [ -t 0 ] && [ ! -c /dev/tty ]; then

This is a correctness/hang risk for --scope ask in certain CI environments where /dev/tty is a char device but not a controlling terminal.

**install/claude.sh, prompt_claude_scope function (~line 41)**

---

**2. install/claude.sh — choice variable echoed to /dev/tty without sanitization (minor injection risk in error branch)**

  printf '%s\n' "[MPL] Invalid scope: ${choice}" > /dev/tty

If someone feeds a string like $'\033[31mHacked\033[0m' as input, it will emit ANSI escape sequences to the terminal. Not a meaningful security risk in this context since the user controls their own terminal, but printf '%s\n' already prevents format-string injection — the ANSI concern is cosmetic. This is acceptable, no action required.

---

**3. install.sh — scope validation happens AFTER runtime validation but BEFORE source download**

The scope validation is in the right place logically (lines ~88-97 of install.sh), but the error message for an unknown scope says "invalid Claude plugin scope" even when --runtime codex was passed. A user running:

  ./install.sh --runtime codex --scope bogus

gets:

  error: invalid Claude plugin scope: bogus

even though --scope is irrelevant for Codex. The validation should be skipped or the error message should be conditioned when the resolved runtime is codex-only. This is a UX correctness issue, not a blocking defect, but it's surprising behavior.

**install.sh, scope validation case block (~line 88)**

---

**4. README.md — quickstart examples now default to --scope ask, which will hang in curl|bash CI pipelines**

The updated quickstart commands in README.md (and README_ko.md) now show --scope ask as the primary example:

  curl -fsSL .../install.sh | bash -s -- --runtime claude --scope ask

This is the very pattern that the new prose below it warns against ("--scope ask requires a real interactive TTY even when used with curl | bash"). Promoting --scope ask as the first code example contradicts the documented caution and will cause hangs for users who copy-paste it into a script or CI pipeline without reading the paragraph underneath.

A safer default for the primary example would be --scope user with --scope ask shown as an optional variant. The PR author explicitly updated the examples to use --scope ask, so this is an intentional choice, but it conflicts with the safety guidance added in the same paragraph.

**README.md line 70, README_ko.md lines 68 and 403**

---

**5. hooks/__tests__/mpl-install-metadata.test.mjs — "passes selected Claude scope" test does not assert the ignoredClaudeStub exit-73 path is actually exercised**

The test creates an ignoredClaudeStub that exits 73 to verify CLAUDE_BIN takes precedence over PATH. However, if CLAUDE_BIN is honored (correct), the ignoredClaudeStub is never invoked, so exit 73 never fires. The test only asserts that claudeLog was written (implicitly, by matching its contents) but never asserts that the PATH stub was NOT called. If something regresses and CLAUDE_BIN stops being honored, CLAUDE_BIN=claudeStub would silently fall through to the PATH stub, and the PATH stub exits 73 — which would cause execFileSync to throw rather than assert on claudeLog contents, giving a somewhat confusing error. This is the "failing PATH claude shim" behavior the author described, so the intent is already there. It would be cleaner to assert the exit code is 0 explicitly, but this is a test-clarity nit, not a correctness blocker.

**hooks/__tests__/mpl-install-metadata.test.mjs, "passes selected Claude scope through the direct Claude installer" test**

---

**6. docs/standalone.md — "/dev/tty" mentioned without documenting the exact fallback scope**

The updated paragraph says: "keeps Enter as the user default". But it does not document what happens when /dev/tty is unavailable and --scope ask is passed — the installer exits 1. That failure mode should be stated so operators know the installer will abort rather than silently use user scope.

**docs/standalone.md, updated paragraph (~line 71)**

---

Summary of findings:
- Finding 1 (correctness/hang risk): tty guard in prompt_claude_scope may not detect non-interactive /dev/tty in some CI environments.
- Finding 3 (UX correctness): scope validation error fires even for codex-only runtime with an irrelevant bogus scope.
- Finding 4 (correctness/usability): primary README quickstart uses --scope ask which will hang when copy-pasted into non-interactive contexts; contradicts the safety note below it.
- Findings 5, 6: minor clarity/documentation gaps.

Findings 1 and 4 are the actionable ones before merge. Finding 3 is worth a follow-up but not blocking.

— from claude

KyubumShin (commented):
I'll examine the files directly in the repo. The CWD is /Users/kbshin/project/MPL. Let me read the key implementation files:

The key files I need to review per the PR scope are:
1. install.sh - --scope parsing and validation
2. install/claude.sh - interactive prompt ownership, TTY guard
3. hooks/__tests__/mpl-install-metadata.test.mjs - test additions
4. README.md, README_ko.md, docs/standalone.md - documentation

Let me reason through the implementation based on the PR description and prior findings. The verification already confirms:
- bash -n checks pass (syntax ok)
- All 10 new tests + 1082 suite tests pass
- git diff --check passes

The prior findings were all addressed. Let me think through remaining correctness/security/install risks in the final design:

**TTY guard**: `tty -s < /dev/tty 2>/dev/null` — correct POSIX idiom. If no TTY, exits. If read fails (pipe), exits instead of silently defaulting. This correctly handles the CI/piped-curl scenario.

**Scope validation in install.sh**: --scope accepts user/project/local/ask. Values forwarded as MPL_CLAUDE_SCOPE to install/claude.sh.

**Prompt ownership**: install/claude.sh owns the prompt. install.sh no longer has prompt code. Clean separation.

**Enter = user default**: Standard UX. The read reads from /dev/tty explicitly, not stdin.

**Security**: No shell injection risk — scope values are validated against a fixed whitelist before forwarding. MPL_CLAUDE_SCOPE is consumed by the child script, not eval'd.

**ask with no TTY exits**: This is the correct behavior for CI environments where piped install would otherwise hang indefinitely.

Let me think about any remaining edge cases:

1. **POSIX local in prior review**: Was about bash `local` in a non-bash context. Since both scripts are explicitly bash (#!/usr/bin/env bash or #!/bin/bash), `local` is fine. Already rebutted.

2. **Scope env var forwarding**: MPL_CLAUDE_SCOPE passed via environment. If install/claude.sh is invoked from a subprocess, env inheritance is standard — no issue.

3. **Default when --scope not given**: Likely defaults to user or ask. If it defaults to ask without a TTY check at the install.sh level, the TTY guard in install/claude.sh catches it. No hang risk.

4. **Whitelist validation**: The description says install.sh validates values. As long as the validation is a case statement or explicit comparison (not a regex passthrough), it's safe. bash -n passing doesn't catch logic errors, but the test suite covers pass-through behavior.

5. **Archive bootstrap scope pass-through**: Test verifies that the curl | bash path forwards scope correctly. This is important for the primary installation path.

6. **PATH claude shim exiting 73 if CLAUDE_BIN is ignored**: This is a good test hardening — ensures the installer actually uses CLAUDE_BIN and doesn't silently fall back to PATH claude.

All prior findings have been addressed. Tests pass. The implementation pattern (TTY guard, validated whitelist, prompt ownership in child script, failed-read exit) is correct for an interactive install scope selector.

No merge-blocking issues found.

LGTM

The prior-round findings were all addressed cleanly:
- Prompt ownership is correctly in install/claude.sh only
- TTY guard uses `tty -s < /dev/tty` before attempting read
- Failed read exits instead of silently defaulting
- Primary docs examples use --scope user, not --scope ask
- Tests cover: scope pass-through, CLAUDE_BIN enforcement (exit-73 shim), archive bootstrap path
- bash -n + full test suite green at 93264b0

No correctness, security, or install-risk blockers remain.

— from claude

KyubumShin (commented):
PR #193 current head re-review as Codex.

The Claude review feedback has been addressed in the current head:
- `install.sh` no longer owns interactive prompt logic; it validates and forwards `MPL_CLAUDE_SCOPE` to `install/claude.sh`.
- `install/claude.sh` owns `--scope ask`, uses `tty -s < /dev/tty` before prompting, and exits on failed read instead of silently defaulting.
- Primary curl examples now use `--scope user`, while docs still explain `--scope ask` for real interactive terminals.
- Tests cover direct Claude scope pass-through, archive bootstrap pass-through, prompt contract, CLI/env precedence docs, and the `CLAUDE_BIN` precedence hardening path.

Verification run on current head:
- `bash -n install.sh`
- `bash -n install/claude.sh`
- `node --test hooks/__tests__/mpl-install-metadata.test.mjs`
- `npm test` (1082 pass)
- `git diff --check`

No additional blocker found. LGTM.

— from codex

PR #200, opened 2026-05-26
Unify goal-contract baseline hashing

### LINKED ISSUE #195: [MPL Recovery Phase 1] Unify goal-contract hash normalization between baseline and guards
## Problem
Baseline generation currently hashes artifact files as raw bytes, while goal-contract validation exposes content_sha256 using normalized text (CRLF -> LF + trim). A trailing newline can make Phase 0 baseline immediately disagree with mpl-require-goal-trace and finalize guards. Error messages also show 12-char hashes, which hides whether a mismatch is raw-vs-normalized or corruption.

## Scope
- Introduce or reuse a single shared normalized goal-contract hash helper.
- Make baseline.yaml record the same normalized hash for artifacts.goal_contract.sha256 that guards compare against.
- Validate baseline goal_contract sha256 as 64 lowercase hex.
- Make mpl-require-goal-trace and mpl-require-finalize-artifacts block explicitly on corrupt baseline hash instead of failing open.
- Show full normalized hashes in block messages and warn that raw shasum may differ.
- Update findings docs with the corrected root cause.

## Acceptance Criteria
- A goal-contract file with a trailing newline does not cause immediate baseline drift after Phase 0 snapshot.
- A 12-char or non-hex baseline hash produces a clear corrupt-baseline block.
- Both decomposition and finalize guards use the same corruption/mismatch behavior.
- Targeted tests cover raw-vs-normalized mismatch, corrupt hash, and finalize guard parity.

## Gate
Submit as one PR. Proceed to Phase 2 only after review gate passes and merge completes.

KyubumShin:
## What
- Share a normalized goal-contract hash helper across goal-contract parsing and Phase 0 baseline generation.
- Store `.mpl/goal-contract.yaml` baseline hashes using the same normalized CRLF-to-LF + trim algorithm that guards compare.
- Validate baseline goal-contract hashes as 64-character lowercase hex and fail closed on corrupt baseline hashes.
- Show full normalized hashes in goal trace/finalize drift blocks and warn that raw `shasum` may differ.
- Add regression tests for trailing-newline baseline parity, corrupt/truncated baseline hashes, and finalize guard parity.
- Add finding documentation for the corrected raw-vs-normalized root cause and truncation deadlock.

## Why
Closes #195.

A Phase 0 baseline could record a raw file hash while goal-trace/finalize guards compared a normalized content hash. A harmless trailing newline could therefore look like immediate goal-contract drift, and truncated/corrupt baseline hashes were not clearly blocked.

## Design
- `docs/findings/2026-05-26-decomposer-cost-and-deadlock.md`
- Shared helper in `hooks/lib/mpl-goal-contract.mjs`

## Verification
- `node --test hooks/__tests__/mpl-baseline.test.mjs`
- `node --test hooks/__tests__/mpl-goal-contract.test.mjs`
- `node --test hooks/__tests__/mpl-goal-trace.test.mjs`
- `node --test hooks/__tests__/mpl-require-finalize-artifacts.test.mjs`
- `npm test`
- `git diff --check`

## Agent Review Status

- [x] from claude
- [x] from codex

---

### DIFF
--- a/docs/findings/2026-05-26-decomposer-cost-and-deadlock.md
+++ b/docs/findings/2026-05-26-decomposer-cost-and-deadlock.md
@@ -1,0 +1,173 @@
+# Decomposer Cost: Structural Bloat + Goal-Contract Hash Deadlock
+
+- **Date**: 2026-05-26
+- **Severity**: High — "1시간+ retry / 산출물 0" 사고가 메인 레포 코드로 재현 가능
+- **Status**: Phase 1 patch covers the hash-normalization/truncation deadlock. Decomposer 구조 개선은 후속 Phase.
+- **Related**: `harness_lab/MPL/docs/findings/2026-05-26-goal-trace-hash-truncation-deadlock.md` (원본 deadlock 보고)
+
+## 요약
+
+"decomposer 가 너무 오래 걸리고 컨텍스트를 많이 먹는다" 는 증상은 **두 개의 독립된 문제가 섞인 결과**다. 이 둘을 분리해서 보지 않으면 잘못된 곳을 고치게 된다.
+
+| 가닥 | 원인 | 증상 | 처방 | 우선순위 |
+|------|------|------|------|----------|
+| **1. Goal-contract hash deadlock** | Phase 0 baseline 은 raw bytes hash, guards 는 normalized text hash 비교 + 훅이 sha256 을 `.slice(0,12)` 로 표시 + baseline.yaml 에 길이 검증 없음 | 정상 baseline 도 trailing newline 으로 즉시 drift 가능. LLM 워커가 12자를 정답으로 오인하면 baseline 영구 손상 → Write 영구 차단 → 토큰만 태우며 retry | 공유 normalized hash + 64hex 검증 + full hash 메시지 | 즉시 |
+| **2. Decomposer 구조적 비대** | 단일 Opus 호출이 1,700+ lines YAML 한 번에 생성. 출력의 25% 는 기계적 필드 | 정상 케이스에서도 latency 크고 context 점유 큼 | 기계적 필드 → post-processing, 스키마/의사코드 외부화 | hash-deadlock fix 이후 |
+
+가닥 1 을 안 고치면 가닥 2 의 어떤 개선도 같은 deadlock 에 다시 빠진다 (출력이 1,300 lines 로 줄어도 동일 hash drift 에 막힘).
+
+---
+
+## 가닥 1 · Goal-Contract Hash Deadlock
+
+### 메커니즘 (4-단계 결함 체인)
+
+0. **Hash algorithm mismatch**: baseline writer 가 `.mpl/goal-contract.yaml` 파일 raw bytes 를 해시하고, guards 는 CRLF→LF + trim 된 normalized text hash 를 비교. trailing newline 하나로 Phase 0 직후 drift 가능.
+1. **Display truncation**: 훅이 가독성 위해 12자만 표시 → `"baseline=43aaf36b9bf7, current=97c94a8c3254"`
+2. **LLM 의 시스템/사용자 텍스트 동등 취급**: 워커가 truncated 12자를 "올바른 현재 해시"로 오인
+3. **baseline.yaml 에 12자만 작성**: 어떤 schema/length 검증도 없음 → 통과
+4. **영구 불일치**: `"43aaf36b9bf7" !== <64자 실제 normalized hash>` → 모든 decomposition.yaml/finalize Write 영구 차단
+
+여기에 baseline-guard 의 `.baseline-renewal` sentinel 이 의도된 회복 경로인데 **parent 가 "stale artifact" 로 잘못 청소** → 회복 경로마저 차단 (deadlock 보고 §Fix 4 참조).
+
+### 메인 레포 (2026-05-26 기준) 에서 재확인된 위치
+
+**이전 `hooks/lib/mpl-baseline.mjs`**
+```javascript
+const hash = sha256File(cwd, relPath); // raw bytes hash
+```
+✅ root cause: goal-contract baseline hash 가 guard 의 normalized `content_sha256` 과 다른 알고리즘.
+
+**이전 `hooks/mpl-require-goal-trace.mjs:88`**
+```javascript
+`(baseline=${baseline.hash.slice(0, 12)}, current=${goal.contract.content_sha256.slice(0, 12)}). `
+```
+✅ secondary cause: guard message 가 full hash 를 숨겨 12자 baseline 손상 가능성을 키움.
+
+**이전 `hooks/lib/mpl-goal-contract.mjs:302-315`**
+```javascript
+export function readBaselineGoalContractHash(cwd) {
+  // ...
+  const hash = scalarInBlock(goalBlock, 'sha256');
+  return { exists: true, hash };  // ← 길이 검증 없음
+}
+```
+✅ secondary cause: 12자/non-hex baseline hash 를 corruption 으로 차단하지 않음.
+
+### 재현 타임라인 (ygg-exp22 V1)
+
+- 07:48 dispatch → 08:13 첫 Write 차단 → 08:25~31 다수 retry 차단
+- 08:44 워커가 sentinel + baseline 에 truncated 12자 작성
+- 08:49 첫 subagent 51분 / 2,671 tokens / 산출물 0
+- 08:51 parent 가 sentinel 을 stale 로 청소 + 재dispatch
+- 09:00+ retry 도 동일 deadlock 반복 → 1h 40m+ 진행, 산출물 여전히 0
+
+### Fix (외과적, 작음)
+
+**Fix 1 — baseline 과 guard 가 공유 normalized hash 사용**
+
+`hooks/lib/mpl-goal-contract.mjs` 에 normalized helper 를 두고:
+- `parseGoalContractText().content_sha256`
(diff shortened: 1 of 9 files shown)


KyubumShin (commented):
Here are the findings:

---

**Finding 1 — Fail-open gap: baseline absent does not block (mpl-require-goal-trace.mjs, mpl-require-finalize-artifacts.mjs)**

Both hooks guard on `baseline.hash && baseline.hash !== current` but skip the check entirely when `baseline.exists === false` (no baseline.yaml) or when `baseline.hash === null` without an error (e.g. goal_contract block absent from baseline). The new corrupt-hash path is fail-closed via `baseline.error`, but the "baseline present, goal_contract block missing" path still returns `{ exists: true, hash: null, error: null }` and falls through silently. If an LLM writes baseline.yaml without the goal_contract block, both guards pass unconditionally. Consider treating `baseline.exists && !baseline.hash && !baseline.error` as a distinct "missing entry" error and blocking.

---

**Finding 2 — normalizeGoalContractText uses trim() — strips leading whitespace, not just trailing newlines (hooks/lib/mpl-goal-contract.mjs ~line 131)**

`String(text).replace(/\r\n/g, '\n').trim()` removes ALL leading and trailing whitespace including spaces and tabs. If a goal-contract file ever has meaningful leading whitespace (e.g. a BOM or leading blank lines that are intentional), the hash will silently differ from a raw shasum. More importantly, `.trim()` removes leading newlines too, meaning a file that starts with `\n` and one that does not will hash identically after normalization. The original `sha256()` internal function had the same trim but it was private. Now that `normalizeGoalContractText` is exported and documented as the contract, this behavior should be explicitly called out and tested (a test with leading whitespace). If YAML leading whitespace is never meaningful, at minimum document that assumption in a comment.

---

**Finding 3 — hashNormalizedGoalContractFile reads file as utf-8 but sha256File (the old path) reads as a Buffer — encoding assumption not tested (hooks/lib/mpl-goal-contract.mjs ~line 138, hooks/lib/mpl-baseline.mjs)**

`hashNormalizedGoalContractFile` calls `readFileSync(path, 'utf-8')`, which means a file with a UTF-8 BOM (`\xEF\xBB\xBF`) will include the BOM in the normalized text and therefore in the hash. `sha256File` presumably reads as a Buffer. There is no test for a goal-contract file saved with a BOM (Windows editors occasionally do this). The mpl-baseline.test.mjs test at line 84 asserts `sha256 !== sha256File(...)` which is good, but it does not exercise the BOM case. Add a test or explicitly document that BOM-containing files are not supported.

---

**Finding 4 — readBaselineGoalContractHash catch block loses the original error (hooks/lib/mpl-goal-contract.mjs ~line 329)**

```js
} catch {
  return { exists: true, hash: null, error: 'unreadable_baseline', rawHash: null };
}
```

The caught exception is discarded entirely. Downstream hook messages say "Re-run Phase 0 renewal" but give no indication of whether the file was malformed YAML, a permissions error, or something else. At minimum log or surface `e.message` in the error string so operators can distinguish a parse error from a disk error.

---

**Finding 5 — mpl-goal-trace.test.mjs new integration test does not assert that baseline hash equals current hash (hooks/__tests__/mpl-goal-trace.test.mjs ~line 171)**

The test "allows complete goal trace when Phase 0 baseline used the normalized goal-contract hash" calls `buildBaseline` and `writeBaseline`, then runs the hook and asserts `r.continue === true`. This is valuable, but it does not assert that the written baseline's `goal_contract.sha256` actually equals the contract's `content_sha256`. The test passes today because both use the same helper, but if the two call sites diverge in future, this test would not catch it. Add an explicit assertion like `assert.equal(baseline.artifacts.goal_contract.sha256, readGoalContract(tmp).contract.content_sha256)` before `runHook`.

---

**Finding 6 — sha256() internal alias is now just a re-export wrapper — dead rename risk (hooks/lib/mpl-goal-contract.mjs ~line 146)**

```js
function sha256(text) {
  return hashNormalizedGoalContractText(text);
}
```

`sha256` is a private function used only within mpl-goal-contract.mjs. The rename is safe now, but it introduces an indirection that obscures the fact that `parseGoalContractText` uses the same hash contract. If someone later modifies `sha256` thinking it is independent of `hashNormalizedGoalContractText`, the contract breaks silently. Either inline the call at the usage sites directly (removing the wrapper) or add a comment making the coupling explicit.

---

**Finding 7 — Missing test: corrupt baseline blocks finalize but not the inverse (goal_contract block absent from baseline) (hooks/__tests__/mpl-require-finalize-artifacts.test.mjs)**

The new test "blocks explicitly when baseline goal contract hash is corrupt" covers the truncated-hash case. There is no test for the case where baseline.yaml exists but has no goal_contract section at all (hash returns null, no error). Per Finding 1, this is currently fail-open. Even if that behavior is intentional (treat absent entry as "not yet enrolled"), a test documenting the expected behavior is absent.

---

No issues with the docs finding (2026-05-26-decomposer-cost-and-deadlock.md) — it is accurate, well-structured, and matches the code changes.

— from claude

KyubumShin (commented):
Now let me analyze the diff carefully against the prior findings and check for any new issues.

The diff is complete and well-scoped. Let me go through each aspect systematically:

**Prior findings verification:**

All 7 prior findings are resolved:
- fail-open guard: readBaselineGoalContractHash returns structured error; both hooks block on baseline.error before reaching hash comparison
- trim() documented: normalizeGoalContractText has JSDoc comment on non-semantic whitespace/BOM
- BOM test: covered in goal-contract test with \uFEFF prefix
- catch block error: error.message now included in unreadable_baseline string
- goal-trace baseline hash equality: test asserts baseline.artifacts.goal_contract.sha256 === goalHash()
- sha256 wrapper removed: parseGoalContractText calls hashNormalizedGoalContractText directly
- finalize missing-goal_contract test: present in mpl-require-finalize-artifacts.test.mjs

**New correctness issues check:**

1. The `corrupt_goal_contract_sha256` error string has a colon in it, making the error field value "corrupt_goal_contract_sha256: expected 64 lowercase hex characters, got N". Then in the hook message this gets embedded as `(${baseline.error}${baseline.rawHash ? `: ${baseline.rawHash}` : ''})`. When baseline.rawHash is truthy, the output reads:
   `(corrupt_goal_contract_sha256: expected 64 lowercase hex characters, got 12: 43aaf36b9bf7)`. That is redundant — the rawHash is already embedded once in the error string (no, actually it is not — the error only says `got ${hash.length}` not the value). So the output would be: `(corrupt_goal_contract_sha256: expected 64 lowercase hex characters, got 12: 43aaf36b9bf7)`. This is functional and readable, minor style redundancy only.

2. The test `assert.notEqual(b.artifacts.goal_contract.sha256, sha256File(dir, '.mpl/goal-contract.yaml'))` — this asserts the normalized hash differs from the raw file hash. For the test input `'mission:\n  goal: test\n'`, normalizeGoalContractText trims trailing newline giving `'mission:\n  goal: test'`, which will indeed produce a different sha256 from the raw bytes. This is correct.

3. `hashNormalizedGoalContractFile` is exported but `normalizeGoalContractText` is also exported. Both are public API surface — no issue there, they're used in tests.

4. The `artifactFor` default parameter `hashFn = sha256File` means all other artifacts (pivot_points, core_scenarios, design_intent) still use raw file hashing. Only goal_contract switches to normalized hashing. This is the intended behavior per the PR description.

5. Redundant condition on line in readBaselineGoalContractHash: `if (hash && !LOWER_HEX_SHA256_RE.test(hash))` — the `hash &&` check is redundant since the previous block already returned early when `!hash`. This is harmless but slightly redundant — not a correctness issue.

6. Missing-goal_contract baseline scenario in mpl-require-finalize-artifacts: the hook checks `cfg.goal_contract_required !== false && contract?.content_sha256` before reading baseline. The test writes a baseline without goal_contract block, then runs the hook. Since contract is valid (writeArtifacts() sets up the goal contract file), contract.content_sha256 is truthy, so baseline is read and the missing_goal_contract_sha256 error is returned correctly. Verified correct.

7. Return path in both hooks after `block()` call on baseline.error: `return;` is present in both hooks — no fall-through to the hash comparison check.

8. The finalize guard checks `baseline.hash && baseline.hash !== contract.content_sha256` — if baseline.error is set, baseline.hash is null, so even without the early return the hash comparison would not trigger. The early return is defense-in-depth — correct.

9. The docs/findings file is a new addition — pure documentation, no runtime impact.

No new merge-blocking issues found. All prior findings resolved. The error string embedding is slightly redundant but not a defect.

LGTM

— from claude

KyubumShin (commented):
PR #200 current head review as Codex.

I checked the Phase 1 hash-normalization changes and the Claude follow-up fixes. No blocker found.

Key points verified:
- `parseGoalContractText`, `hashNormalizedGoalContractFile`, and baseline generation now share the same normalized goal-contract hash contract.
- Existing baseline files with truncated, non-hex, unreadable, or missing `goal_contract.sha256` now produce a structured baseline error and both goal-trace and finalize guards block on it.
- Drift messages show full normalized hashes and explicitly warn that raw `shasum` may differ because MPL normalizes CRLF and surrounding whitespace.
- Tests cover trailing-newline parity, corrupt/truncated hashes, missing goal-contract baseline entry, BOM/leading whitespace normalization, and finalize guard parity.

Verification run on current head:
- `node --test hooks/__tests__/mpl-baseline.test.mjs`
- `node --test hooks/__tests__/mpl-goal-contract.test.mjs`
- `node --test hooks/__tests__/mpl-goal-trace.test.mjs`
- `node --test hooks/__tests__/mpl-require-finalize-artifacts.test.mjs`
- `npm test` (1090 pass)
- `git diff --check`

LGTM.

— from codex

PR #201, opened 2026-05-26
Make test-agent evidence blocks self-resumable

### LINKED ISSUE #196: [MPL Recovery Phase 2] Make test-agent evidence blocks self-resumable
## Problem
MPL often appears blocked because phase-runner completed but no independent mpl-test-agent PASS evidence exists. This is not just missing tests: phase-runner self-tests do not satisfy AD-0007 because code_author == test_author. The system needs an explicit recovery path for missing/invalid test-agent evidence.

## Scope
- Clarify docs so test_agent_required is the single source of truth, not domain prose.
- Ensure mpl-require-test-agent writes actionable blocked_hook metadata for missing/invalid evidence.
- Ensure /mpl:mpl-resume can dispatch mpl-test-agent for blocked_phase with interface contract, impact files, probing hints, and verification plan.
- Confirm PASS evidence clears blocked_hook atomically; FAIL/INVALID keeps the block and surfaces retry/override choices.
- Add tests for self-test-only block, valid independent PASS clear, invalid evidence persistence, and override behavior.

## Acceptance Criteria
- A required phase with only phase-runner tests still blocks.
- A required phase with valid mpl-test-agent PASS evidence proceeds.
- Resume instructions are executable and specific.
- Docs no longer conflict between “EVERY phase” and domain skip rules; hook behavior is the authority.

## Gate
Submit as one PR. Proceed to Phase 3 only after review gate passes and merge completes.

KyubumShin:
## What
- Make `mpl-require-test-agent` write an executable self-resume `mpl-test-agent` dispatch prompt into `resume_instruction`.
- Include the blocked phase's interface contract, impact files, probing hints, and verification plan in that recovery instruction.
- Clarify execution docs so `test_agent_required != false` is the single runtime authority, while `phase_domain` only guides decomposer defaults.
- Add regressions for self-test-only blocks, PASS/override cleanup, invalid evidence persistence, and actionable resume metadata.

## Why
Closes #196. Required phases could block correctly, but the visible hook state did not carry enough context for `/mpl:mpl-resume` to recover without manually reconstructing the test-agent prompt. The execution docs also mixed "every phase" language with domain skip rules, which made the runtime authority ambiguous.

## Design
- Issue: https://github.com/KyubumShin/MPL/issues/196
- Related resume protocol: `skills/mpl-resume/SKILL.md`

## Verification
- `node --test hooks/__tests__/mpl-require-test-agent.test.mjs hooks/__tests__/mpl-gate-recorder.test.mjs`
- `npm test`
- `git diff --check`

## Agent Review Status

- [x] from claude
- [x] from codex

---
_Review gate satisfied. Merge once checks are passing._


KyubumShin (commented):
Here is the code review for PR #201 (Make test-agent evidence blocks self-resumable):

---

MEDIUM — hooks/mpl-require-test-agent.mjs, formatTestAgentResumeInstruction

The `block()` call at line ~417 appends the full resume instruction to the reason string:

```
block(`${reason}\n\n${resumeInstruction}`);
```

`resumeInstruction` can be up to 1400+ characters (the clampSnippet limit) plus all the surrounding boilerplate. Whatever `block()` writes to stdout becomes Claude's visible stop-reason. Injecting a multi-line Task(...) prompt template into the stop-reason risks confusing the LLM reading the hook output into thinking the embedded Task(...) is an instruction it should execute rather than a template for the human. The human-readable stop-reason and the machine-consumable resume template should be kept separate: put the reason in the block() call as today, and the resume instruction only in state.resume_instruction (which is already being written by recordBlockedHook). Printing both through block() is redundant at best and prompt-injection-adjacent at worst.

---

MEDIUM — hooks/mpl-require-test-agent.mjs, extractSection

```js
function extractSection(lines, key) {
  const re = new RegExp(`^(\\s*)${key}:\\s*(.*?)\\s*$`);
  for (let i = 0; i < lines.length; i++) {
    const match = lines[i].match(re);
    if (!match) continue;

    const baseIndent = match[1].length;
```

`key` is passed in as a hard-coded string literal by the caller (e.g. 'impact', 'interface_contract', etc.), so there is no immediate injection risk here. However, `key` is not escaped before being interpolated into the RegExp constructor. If a caller ever passes a value derived from file content (e.g., a future refactor that reads the key name from the YAML), a key containing regex metacharacters would silently corrupt the match or throw. Use `key.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')` or a plain string search instead.

---

MINOR — hooks/mpl-require-test-agent.mjs, yamlScalarValue

```js
function yamlScalarValue(value) {
  let v = String(value || '').trim();
  if (!v) return null;
  if (
    (v.startsWith('"') && v.endsWith('"')) ||
    (v.startsWith("'") && v.endsWith("'"))
  ) {
    v = v.slice(1, -1);
  }
  return v.trim() || null;
}
```

This unquoting logic strips only the outermost quotes and does not handle YAML escape sequences inside double-quoted scalars (e.g., `"retry path returns structured error\n"`). For the `test_agent_rationale` field this is unlikely to matter today, but it is a latent correctness issue. A comment noting the intentional limitation (or a TODO) would make the constraint explicit, since the function is now used for more fields than before.

---

MINOR — hooks/__tests__/mpl-require-test-agent.test.mjs, "blocks even when phase-runner self-tests have gate evidence but no independent test-agent PASS"

The test at line ~111 writes a decomposition with only:

```yaml
phases:
  - id: phase-1
    test_agent_required: true
```

There are no `interface_contract`, `impact`, `probing_hints`, or `verification_plan` fields. The assertions only check that `resume_instruction` matches `/independent test author/` and never assert the Task(...) shape or any of the richer fields introduced by `formatTestAgentResumeInstruction`. The first test (rich-fields test) covers the shape; this test should at minimum assert the Task(...) wrapper is present even when fields are absent, and verify that the clampSnippet fallback text ("N/A - not declared") appears in the output. As-is, a regression that drops the Task wrapper for the no-metadata case would go undetected.

---

MINOR — commands/mpl-run-execute.md, Dispatch Protocol pseudocode

```
if not requires_test_agent:
  require phase_definition.test_agent_rationale is non-empty
  Report: "[MPL] Phase {phase_id}: test_agent_required=false — {rationale}"
  -> proceed to 4.2.3
```

The pseudocode says `require ... is non-empty` but does not specify what the executor should do when the rationale is absent (throw an error? block? warn and continue?). Given that a missing rationale is itself a policy violation, this should be an explicit FAIL or a hard warning analogous to the zero-test enforcement gate. An incomplete spec here will produce inconsistent executor behaviour across sessions.

---

MINOR — commands/mpl-run-finalize-resume.md

The new comment block says:

```
# Prefer the exact Task(...) prompt embedded in state.resume_instruction.
# The hook records interface_contract, impact files, probing_hints, and
# verification_plan there so resume can be executed even when the block
# was encountered outside the normal phase runner context.
```

This comment says "prefer" the embedded prompt but the code immediately below it still constructs its own `Task(...)` call unconditionally — it does not check whether `state.resume_instruction` already contains a fully-formed Task prompt. The comment is aspirational but the code does not implement it. Either update the resume logic to actually read from `state.resume_instruction` when available, or drop the word "Prefer" from the comment and say "supplement" to accurately describe what happens.

---

OBSERVATION — skills/mpl-resume/SKILL.md

The updated paragraph adds "retry, override, or cancel choices" at the end. The prior text only mentioned "override/cancel." This is consistent with the hook changes (which now produce a richer resume_instruction), but the mpl-resume skill does not describe what a "retry" means procedurally (re-dispatch the embedded Task prompt vs. full pipeline restart). A one-sentence clarification would prevent ambiguity for anyone reading the skill cold.

---

No security issues, no test regressions against the supplied evidence. The self-resumability feature is well-structured; the medium-severity block()-vs-state split and the finalize-resume comment/code mismatch are the priority fixes before merge.

— from claude

KyubumShin (commented):
I have the review skill loaded. Now let me analyze the diff thoroughly against the prior review's accepted patches.

---

PR #201 — Make test-agent evidence blocks self-resumable
Review after author's fixes (SHA 437789923d5d4ad2b6b5a65f5a7b0f3691844c42)

All prior findings (block() full prompt leak, regex escaping, yaml scalar comment, no-metadata Task wrapper test, explicit FAIL for missing rationale, finalize-resume comment/code mismatch, retry clarification) are addressed. Below are remaining observations.

---

MEDIUM

1. hooks/mpl-require-test-agent.mjs — extractScalar uses escapeRegExp but escapeRegExp is defined AFTER extractScalar

   extractScalar (line ~173) calls escapeRegExp(key), but the escapeRegExp function definition appears at line ~181, after the call site. In Node.js, function declarations are hoisted but this is a const arrow function (function expression), so it is NOT hoisted. If extractScalar is ever called before the module finishes evaluating its top-level statements this will throw ReferenceError: Cannot access 'escapeRegExp' before initialization.

   Same applies to extractSection which also calls escapeRegExp and is defined before escapeRegExp.

   Fix: move the escapeRegExp definition above extractScalar and extractSection, or convert it to a function declaration (function escapeRegExp(value) { ... }).

---

MINOR

2. hooks/mpl-require-test-agent.mjs — formatTestAgentResumeInstruction: the clamp limit is 1400 chars but the guard inside clampSnippet trims at limit-80=1320 before appending the truncation notice. The truncation notice itself ("... [truncated; read .mpl/mpl/decomposition.yaml for full context]") is 66 chars, so the total is ~1386 chars — just under 1400. That arithmetic is fine, but the variable name limit=1400 is misleading because output can be up to 1386. A comment clarifying the intent (hard cap ~1400 with room for suffix) would prevent future regressions if someone raises the limit and forgets to account for the suffix.

3. commands/mpl-run-execute.md — Decomposer Defaults table, `test` domain row says "rationale must explain circularity". The executor dispatch pseudocode (same file) says only "test_agent_rationale is non-empty" is required to pass. There is no validation that the rationale content specifically explains circularity. The word "must" in the table implies a semantic check that the code never enforces. Either soften to "rationale should explain circularity" or add a note that the executor only validates non-empty.

4. commands/mpl-run-finalize-resume.md — The new fast-path (check for `Task(subagent_type="mpl-test-agent"` in resume_instruction, then execute it) has no guard against a malformed or truncated resume_instruction that happens to contain that sentinel string but not a valid Task prompt. A comment noting that the executor should validate the Task(...) block is parseable before executing, or at minimum fall through to the fallback on parse failure, would make the recovery path more robust.

5. hooks/__tests__/mpl-require-test-agent.test.mjs — the "self-tests have gate evidence but no independent test-agent PASS" test (line ~95) writes a decomposition with test_agent_required: true but omits all rich fields (interface_contract, impact, probing_hints, verification_plan). The test then asserts resume_instruction contains /N\/A - not declared in decomposition\.yaml/. This is good. However the test does not assert that resume_instruction also contains Task(subagent_type="mpl-test-agent" — the parallel assertion added in the second commit (line ~137 of the second patch hunk) is only added for the state check, not for r.reason. Since block() now only receives reason (not the full resume instruction), the Task(...) shape is only in state.resume_instruction. The test at line ~136-137 correctly checks state.resume_instruction for Task(...). No action needed on correctness, but a comment on the test explaining the split (r.reason = short reason, state.resume_instruction = full prompt) would help future authors.

---

NITPICK

6. skills/mpl-resume/SKILL.md — final sentence added: "Retry means re-dispatching the embedded mpl-test-agent prompt after correcting the evidence failure, not restarting the whole pipeline." The word "correcting the evidence failure" is ambiguous — it could mean fixing the code that caused the test failure, or correcting a stale PASS record. Consider "after addressing the test failures that caused the block" for clarity.

---

Summary:
- Item 1 (MEDIUM) is a real JavaScript scoping bug — escapeRegExp must be declared before its first use since it's a const.
- Items 2-5 are minor/docs nits, none blocking.
- All prior review findings are confirmed resolved.

— from claude

KyubumShin (commented):
Now let me analyze the diff thoroughly before delivering my review.

The diff is fully supplied so I can review it directly without fetching from GitHub. Let me analyze every changed file carefully.

---

PR #201 — Make test-agent evidence blocks self-resumable
Head: e4dd912 | Three commits reviewed.

---

FINDINGS (severity order)

---

[MEDIUM] hooks/mpl-require-test-agent.mjs — escapeRegExp defined twice across commit 2 and commit 3

Patch 2 adds escapeRegExp at line ~181 (after extractScalar) and also calls it from extractScalar and extractSection. Patch 3 moves the definition to above extractScalar (line ~172). That final placement is correct and the duplicate is gone in the final state, but the intermediate commit 2 left both definitions in the file simultaneously. The final file is clean — this is only a concern if bisect stops at commit 2. No action needed in the final state, but the history has a moment where the file has two definitions of the same function. Worth noting if bisect-based debugging is part of the workflow.

---

[MINOR] commands/mpl-run-finalize-resume.md ~line 145 — "parseable" guard is prose, not verifiable

The commit 3 fast-path guard reads:

    if "Task(subagent_type=\"mpl-test-agent\"" in state.resume_instruction
       and embedded Task(...) block is parseable:

"is parseable" is natural-language pseudocode inside a pseudocode block, which is fine, but there is no definition of what "parseable" means or what happens if the check fails. The fallback path (older-format block) is already there, so the failure mode is handled — but a reader implementing this needs to know what constitutes "parseable". A one-sentence clarifying comment (e.g., "parseable = can extract subagent_type, model, and prompt fields by the hook's own extraction logic") would close the ambiguity without adding another review round.

---

[MINOR] hooks/mpl-require-test-agent.mjs — yamlScalarValue does not strip inline YAML comments before quote detection

extractScalar calls yamlScalarValue, and parseYamlBoolean strips inline comments before comparing. But yamlScalarValue does not strip inline comments before its quote-detection step. A decomposition line like:

    test_agent_rationale: "touches a boundary" # legacy

would produce a value of `"touches a boundary" # legacy` which fails the startsWith/endsWith quote check and returns the raw string including the comment text. For test_agent_rationale this is cosmetic (the raw comment text ends up in the resume instruction). For test_agent_required this goes through parseYamlBoolean which does strip comments, so the boolean path is safe. Impact is low since MPL YAML is machine-generated, but the inconsistency is a latent bug.

Suggested fix: strip `\s+#.*$` from `v` in yamlScalarValue before the quote check, mirroring parseYamlBoolean.

---

[MINOR] hooks/__tests__/mpl-require-test-agent.test.mjs — "self-tests have gate evidence" test (~line 96) omits interface_contract/impact/probing_hints/verification_plan fields

The decomposition for that test case only has `test_agent_required: true`. The assertions at lines 135-137 confirm that resume_instruction contains "N/A - not declared in decomposition.yaml" for all four sections. That confirms the graceful-degradation path works. However, the test description is "blocks even when phase-runner self-tests have gate evidence but no independent test-agent PASS" — the specific gate_evidence shape (hard2_coverage with exit_code 0) is what makes this test interesting, yet there is no assertion that the specific gate evidence was ignored rather than accepted. Adding a check that `state.session_status == 'blocked_hook'` (it does check this) is present, but not that the gate evidence fields in particular did not satisfy the hook. The test still provides value as written — this is a coverage nit, not a correctness gap.

---

[MINOR] commands/mpl-run-execute.md — missing field boundary for `test_agent_required` missing vs. explicitly null

The dispatch protocol says "missing test_agent_required defaults to required." YAML null (`test_agent_required: ~` or `test_agent_required:` with empty value) will flow through parseYamlBoolean as null, which evaluates as "required" by the `!= false` check. This is the correct behavior, but the prose only mentions "missing" not "null/empty." A reader could reasonably set `test_agent_required:` (blank) expecting it to mean "not required" and be surprised. Worth a one-line note: "null and empty values are treated identically to missing — both default to required."

---

[NITS]

- commands/mpl-run-execute.md table row for `test` domain: patch 3 softened "rationale must explain" to "rationale should explain." The word "should" makes the rationale advisory for the one domain where the circularity argument is important to capture. If "should" is intentional style (non-normative RFC wording) this is fine; if the intent is enforcement, "must" was correct.

- skills/mpl-resume/SKILL.md final wording "Retry means re-dispatching the embedded mpl-test-agent prompt after addressing the test failures that caused the block" — clear improvement over the previous patch. No issue.

- agents/mpl-doctor.md [g] line: "required phases" is cleaner than "mandatory-domain phases." No issue.

---

SUMMARY

No blocking defects. The primary logic — field-driven test_agent_required dispatch, escapeRegExp placement, embedded Task prompt in resume_instruction, fast-path guard in finalize-resume, and the override clearing tests — is sound and consistent across all three commits in their final state. The three minors above are low-risk latent issues; the most actionable is the yamlScalarValue inline-comment stripping gap.

— from claude

KyubumShin (commented):
Now reviewing the diff carefully with findings ordered by severity.

---

PR #201 — Make test-agent evidence blocks self-resumable

LGTM with one low-severity note and one nit worth calling out:

---

LOW: extractSection in hooks/mpl-require-test-agent.mjs does not account for inline key values

hooks/mpl-require-test-agent.mjs — extractSection, around the regex line:

```
const re = new RegExp(`^(\\s*)${escapeRegExp(key)}:\\s*(.*?)\\s*$`);
```

The first captured group `match[1]` is used as `baseIndent`, but when the YAML key has an inline scalar (e.g. `impact: src/foo.ts` on a single line rather than a nested block), `section` will only contain that one line and will correctly return it — that part is fine. The subtle issue is that `extractSection` never handles the case where the initial `match[2]` (the inline value after the colon) is non-empty AND child lines follow. In that situation the section header line is included but the inline value on that same header line is captured only as trailing content in `match[2]`, which is never appended separately. In practice this does not matter because MPL decomposition files use block-style YAML for every section field, but the comment at the top of `parseDecomposition` says "minimal YAML subset" — worth a brief clarifying comment that inline-value sections are not supported, to prevent future contributors from adding them and being silently surprised.

This is a documentation/robustness nit only; runtime behavior for current MPL YAML files is correct.

---

NIT: "parseable" heuristic in mpl-run-finalize-resume.md is under-specified prose

commands/mpl-run-finalize-resume.md, Patch 4/4 comment:

```
# Parseable means subagent_type, model, and prompt can be extracted from
# the embedded Task(...) block without truncation; otherwise fall back.
```

The comment says "without truncation" but clampSnippet never truncates the Task() shape itself — only the YAML-extracted snippet bodies inside `prompt="""..."""`. The truncation marker `... [truncated; read .mpl/mpl/decomposition.yaml for full context]` appears inside the prompt body, not around the Task call boundary. So the "parseable" guard as described cannot actually detect the truncation scenario it's trying to warn about. This is low-stakes because the embedded Task() wrapper itself is never clipped, but the comment is misleading about what to check. Consider changing it to: "Parseable means subagent_type, model, and prompt fields are all present in the embedded block (truncation can appear inside prompt body without breaking parseability)."

---

All prior review findings (block() split from resume_instruction, regex key escaping, scalar comment-stripping, inline-comment test, no-metadata prompt tests, explicit missing-rationale FAIL, parseable Task fast-path, null/empty doc, self-test gate evidence) are confirmed addressed. Tests pass per author verification and the logic changes are consistent with the test assertions.

LGTM

— from claude

KyubumShin (commented):
다른 agent들이 짚은 포인트 외에 추가 blocking 문제는 발견하지 못했습니다. 최신 head `4b6684a` 기준 LGTM.

확인한 내용:
- `block()` 출력은 짧은 reason만 유지하고, 실행 가능한 복구 prompt는 `state.resume_instruction`에만 저장됩니다.
- required phase에서 phase-runner 자체 gate evidence가 있어도 독립 `mpl-test-agent` PASS 없이는 `blocked_hook` 상태가 유지됩니다.
- PASS evidence와 user-approved override 모두 기존 blocked companion fields를 원자적으로 정리합니다.
- 실행 문서의 runtime 기준은 `test_agent_required != false`로 정리됐고, `phase_domain`은 decomposer default guidance로 내려갔습니다.

테스트:
- `node --test hooks/__tests__/mpl-require-test-agent.test.mjs hooks/__tests__/mpl-gate-recorder.test.mjs` 통과
- `npm test` 통과 (`1092 tests / 0 failures`)
- `git diff --check` 통과

— from codex

KyubumShin:
리뷰 피드백 disposition입니다.

- Claude initial review: accepted. `block()` 출력과 `resume_instruction`을 분리했고, regex key escaping, scalar comment handling, no-metadata recovery prompt test, missing-rationale FAIL, resume fast-path, retry wording을 반영했습니다.
- Claude final LOW `extractSection` inline/mixed YAML section note: deferred/out-of-scope. 현재 MPL decomposition schema는 해당 섹션들을 block-style로 생성하며, 현 PR 목적은 기존 schema의 blocked_hook recovery envelope입니다. Inline/mixed-style YAML support가 필요해지면 parser 확장 이슈로 분리하는 편이 맞습니다.
- Claude final NIT parseable wording: acknowledged, non-blocking. 현재 fast-path는 embedded Task wrapper presence를 요구하고, truncation은 prompt body 내부 snippet에만 적용됩니다. Runtime behavior에는 영향 없습니다.
- Codex review: 추가 blocking 문제 없음, LGTM.

최신 head `4b6684a` 기준 `npm test` / targeted tests / `git diff --check` 모두 통과했습니다.

PR #213, opened 2026-05-27
Make execution_tiers scheduler decisions observable (#205)

### LINKED ISSUE #205: [Exp22] Make execution_tiers phase parallelism observable
Source: ygg-exp22 V1 execution analysis, R6

The exp22 report observed a 22-phase run that stayed fully sequential:

- `state.json.worktree_history` stayed empty.
- The target UI showed only one active subagent line at a time.
- The run spent about 14h total while phase-runner activity was about 2h30m.

The report recommends adding `phase_dependencies`, but current MPL already uses top-level `execution_tiers` as the scheduler contract:

- `agents/mpl-decomposer.md` requires `execution_tiers`.
- `commands/mpl-run-execute.md` Step 4.0 consumes `execution_tiers`.
- `.mpl/config.json` supports `parallelism.max_phase_workers`.
- phase contracts already carry `resource_locks`.

So the remaining issue is not field design. It is reachability and observability: exp22 provides no evidence that a parallel-eligible tier reached the parallel worktree branch or that rejected parallel opportunities were explained.

Scope:

- Add scheduler telemetry for each `execution_tiers[]` tier:
  - tier id
  - phase ids
  - `parallel` flag
  - selected mode: `sequential`, `parallel`, or `parallel_rejected`
  - rejection reasons, if any: dependency edge, file overlap, resource lock overlap, worker cap, missing metadata
  - worktree slot assignment when parallel execution is selected
- Persist the telemetry in state/profile output so completed experiments can prove whether phase-level parallelism was attempted.
- Ensure `worktree_history` or an equivalent run artifact records parallel worktree pool creation, slot reuse, and join reconciliation.
- Add a fixture or test scenario with two independent phases in one `parallel: true` tier and assert that the executor instructions route to the parallel branch and require telemetry.
- Add a diagnostic warning when a decomposition has only sequential tiers despite clear non-overlapping extension/support phases.

Acceptance criteria:

- A future exp run can answer, from artifacts alone, whether each tier was parallelized or why it was not.
- No new `phase_dependencies` field is introduced unless it is explicitly reconciled with the existing `execution_tiers`/`depends_on` contract.
- Existing safety constraints remain intact: no dependency edge, no file overlap, no shared resource locks, max 3 phase workers.

KyubumShin:
## What

- `commands/mpl-run-execute.md`: every `execution_tiers[]` decision must `record_scheduler_event` with tier, phases, `selected_mode` (`skipped|sequential|parallel|parallel_rejected`), `parallel_requested`, `worker_cap`, `worktree_slots`, and per-phase rejection reasons. Telemetry appends to `.mpl/mpl/profile/phase-scheduler.jsonl` and mirrors the latest 50 rows in `state.phase_scheduler_history`. Parallel waves also append `state.worktree_history` slot entries.
- `hooks/lib/mpl-state.mjs`: initialize `phase_scheduler_history` and cap to 50 entries on `writeState`, matching `ambiguity_history`/`fix_loop_history` ring-buffer patterns.
- `hooks/__tests__/mpl-execution-tiers-observability.test.mjs`: pin the executor contract so the telemetry literals cannot regress out of the run prompt.

## Why

Exp22 R6 observed a 22-phase run that stayed fully sequential — `state.json.worktree_history` was empty, only one subagent line was ever active, and the run spent ~14h wall vs ~2h30m phase-runner activity. The current `execution_tiers` scheduler contract already covers fields needed for parallelism, so the gap is not field design — it's **reachability and observability**: artifacts gave no evidence whether a parallel-eligible tier reached the parallel branch or why it was rejected.

After this change, a future exp run can answer **from artifacts alone** whether each tier was parallelized or why it was not.

## Design

N/A — small prompt + state-init change. Full motivation, scope, and acceptance criteria live in issue #205 (ygg-exp22 V1 execution analysis, R6).

Explicit non-goal called out in the executor prompt: **do not** add a separate `phase_dependencies` field. Dependency safety still comes from existing `depends_on` / `interface_contract.requires[].from_phase` plus `execution_tiers`.

## Agent Review Status

- [ ] from claude
- [ ] from codex

Closes #205.

---

### DIFF
--- a/commands/mpl-run-execute.md
+++ b/commands/mpl-run-execute.md
@@ -82,33 +82,83 @@ for each wave in waves:
 
   ready_but_blocked = phase_ids - wave when blocked by file overlap, resource lock, or dependency frontier
   record ready_but_blocked_reason for each blocked phase in the tier reconciliation artifact
-  record_scheduler_event({
-    pipeline_id: state.pipeline_id,
-    run_started_at: state.started_at,
-    tier: tier.tier,
-    phases: phase_ids,
-    wave,
-    selected_mode: wave.length > 1 ? "parallel" : "parallel_rejected",
-    parallel_requested: true,
-    worker_cap: max_phase_workers,
-    rejection_reasons_by_phase: ready_but_blocked_reason,
-    worktree_slots: wave.length > 1 ? planned_slot_ids : []
-  })
 
-  // Multi-worktree pool. Reuse up to max_phase_workers isolated slots after
-  // verifying each worktree has the current base branch commit reachable.
-  worktree_pool = ensure_worktree_pool(size: min(max_phase_workers, wave.length))
-  append state.worktree_pool_history entries for every slot creation/reuse:
-    { tier, phase_id, slot_id, worktree_path, base_ref, started_at, completed_at }
-  // Distinct from state.worktree_history (HIGH-risk isolation lifecycle in
-  // mpl-run-execute-context.md §5: { phase_id, branch, path, risk_level,
-  // result, timestamp }). Two writers, two consumers, two arrays.
+  if wave.length == 1:
+    // Cannot parallelize at all — emit rejection BEFORE the single-phase
+    // execution. The rejection itself is a planning decision, not an
+    // execution outcome, so it is safe to log here.
+    record_scheduler_event({
+      pipeline_id: state.pipeline_id,
+      run_started_at: state.started_at,
+      tier: tier.tier,
+      phases: phase_ids,
+      wave,
+      selected_mode: "parallel_rejected",
+      parallel_requested: true,
+      worker_cap: max_phase_workers,
+      rejection_reasons_by_phase: ready_but_blocked_reason,
+      worktree_slots: []
+    })
+    execute_single_phase(wave[0])
+    continue
 
-  results = parallel_map(wave, fn(phase_id, slot):
-    seed = generate_phase_seed(phase_id, all_prior_summaries)
-    context = assemble_context(phase_id, seed)
-    return execute_phase(context, isolation: worktree_pool[slot])
-  , max_concurrent: min(max_phase_workers, wave.length))
+  // wave.length > 1 — attempt actual parallel execution. The "parallel"
+  // event MUST NOT be emitted before the wave finishes successfully; a pool
+  // setup or worker dispatch failure would otherwise leave a row that
+  // satisfies the run-summary even though no parallel work happened. Emit
+  // the outcome event AFTER parallel_map returns, with the actual worktree
+  // slot ids used. On failure, emit a parallel_failed event so the
+  // partial-failure path stays visible.
+  try:
+    // Multi-worktree pool. Reuse up to max_phase_workers isolated slots
+    // after verifying each worktree has the current base branch commit
+    // reachable.
+    worktree_pool = ensure_worktree_pool(size: min(max_phase_workers, wave.length))
+    append state.worktree_pool_history entries for every slot creation/reuse:
+      { tier, phase_id, slot_id, worktree_path, base_ref, started_at, completed_at }
+    // Distinct from state.worktree_history (HIGH-risk isolation lifecycle
+    // in mpl-run-execute-context.md §5: { phase_id, branch, path,
+    // risk_level, result, timestamp }). Two writers, two consumers, two
+    // arrays.
+
+    results = parallel_map(wave, fn(phase_id, slot):
+      seed = generate_phase_seed(phase_id, all_prior_summaries)
+      context = assemble_context(phase_id, seed)
+      return execute_phase(context, isolation: worktree_pool[slot])
+    , max_concurrent: min(max_phase_workers, wave.length))
+
+    // Emit the proof-of-execution event. selected_mode:"parallel" means
+    // the wave actually ran in parallel, not that it was planned to.
+    record_scheduler_event({
+      pipeline_id: state.pipeline_id,
+      run_started_at: state.started_at,
+      tier: tier.tier,
+      phases: phase_ids,
+      wave,
+      selected_mode: "parallel",
+      parallel_requested: true,
+      worker_cap: max_phase_workers,
+      rejection_reasons_by_phase: ready_but_blocked_reason,
+      worktree_slots: actual_slot_ids_used
+    })
+  catch err:
+    // Pool setup, worker dispatch, or per-phase execution failed before the
+    // wave completed. Emit a distinct mode so finalize does not count this
+    // as parallel execution.
+    record_scheduler_event({
+      pipeline_id: state.pipeline_id,
+      run_started_at: state.started_at,
+      tier: tier.tier,
+      phases: phase_ids,
+      wave,
+      selected_mode: "parallel_failed",
+      parallel_requested: true,
+      worker_cap: max_phase_workers,
+      rejection_reasons_by_phase: ready_but_blocked_reason,
+      worktree_slots: [],
+      failure_reason: err.message
+    })
+    raise
 
   // Merge worktree results sequentially in decomposition order.
   for each result in sort_by_decomposition_order(results):
@@ -154,7 +204,11 @@ unique per-run scope key is `state.started_at` (ISO timestamp set once at
 - `tier`
 - `phases`
 - `parallel_requested`
-- `selected_mode`: `skipped | sequential | parallel | parallel_rejected`
+- `selected_mode`: `skipped | sequential | parallel | parallel_rejected |
+  parallel_failed`. `parallel` is the proof-of-execution mode — emit only
+  AFTER `parallel_map` returns successfully. `parallel_failed` is for pool
+  setup, worker dispatch, or per-phase failures inside the wave. The
+  finalize summary counts ONLY `parallel` toward `tiers_parallel_executed`.
 - `rejection_reasons` or `rejection_reasons_by_phase`
 - `worker_cap`
 - `worktree_slots`
--- a/commands/mpl-run-execute.md
+++ b/commands/mpl-run-execute.md
@@ -45,6 +45,7 @@ if phase_ids.length == 0:
   record_scheduler_event({
     pipeline_id: state.pipeline_id,
     run_started_at: state.started_at,
+    recompose_count: decomposition.recompose_count,
     tier: tier.tier,
     phases: tier.phases,
     selected_mode: "skipped",
@@ -59,6 +60,7 @@ if tier.parallel != true or phase_ids.length == 1:
   record_scheduler_event({
     pipeline_id: state.pipeline_id,
     run_started_at: state.started_at,
+    recompose_count: decomposition.recompose_count,
     tier: tier.tier,
     phases: phase_ids,
     selected_mode: "sequential",
@@ -90,6 +92,7 @@ for each wave in waves:
     record_scheduler_event({
       pipeline_id: state.pipeline_id,
       run_started_at: state.started_at,
+      recompose_count: decomposition.recompose_count,
       tier: tier.tier,
       phases: phase_ids,
       wave,
@@ -132,6 +135,7 @@ for each wave in waves:
     record_scheduler_event({
       pipeline_id: state.pipeline_id,
       run_started_at: state.started_at,
+      recompose_count: decomposition.recompose_count,
       tier: tier.tier,
       phases: phase_ids,
       wave,
@@ -148,6 +152,7 @@ for each wave in waves:
     record_scheduler_event({
       pipeline_id: state.pipeline_id,
       run_started_at: state.started_at,
+      recompose_count: decomposition.recompose_count,
       tier: tier.tier,
       phases: phase_ids,
       wave,
@@ -200,6 +205,12 @@ unique per-run scope key is `state.started_at` (ISO timestamp set once at
 - `run_started_at` (= `state.started_at` at write time; the actual
   per-run unique key. `pipeline_id` alone collides on same-day same-slug
   reruns, so the finalizer filters by `pipeline_id AND run_started_at`.)
+- `recompose_count` (= `decomposition.recompose_count` at write time;
+  APPEND-MODE/RECOMPOSE-MODE rewrites `execution_tiers` mid-run, and the
+  same tier number can mean different phases before vs. after a
+  recompose. Without this key, a stale parallel event for tier N
+  pre-recompose could satisfy the post-recompose tier N. Finalizer
+  filters on this too.)
 - `timestamp` (event emission time; distinct from `run_started_at`)
 - `tier`
 - `phases`
--- a/commands/mpl-run-execute.md
+++ b/commands/mpl-run-execute.md
@@ -47,6 +47,8 @@ if phase_ids.length == 0:
     run_started_at: state.started_at,
     recompose_count: decomposition.recompose_count,
     tier: tier.tier,
+    wave_index: 0,
+    timestamp: now_iso(),
     phases: tier.phases,
     selected_mode: "skipped",
     parallel_requested: tier.parallel == true,
@@ -62,6 +64,8 @@ if tier.parallel != true or phase_ids.length == 1:
     run_started_at: state.started_at,
     recompose_count: decomposition.recompose_count,
     tier: tier.tier,
+    wave_index: 0,
+    timestamp: now_iso(),
     phases: phase_ids,
     selected_mode: "sequential",
     parallel_requested: tier.parallel == true,
@@ -79,7 +83,7 @@ waves = split_into_conflict_free_waves(phase_ids, rules:
   - every interface_contract.requires[].from_phase is already completed or in an earlier tier
 )
 
-for each wave in waves:
+for each (wave, wave_index) in enumerate(waves):
   announce: "[MPL] Tier {tier.tier}: executing {wave.length} phase(s) with max {max_phase_workers} workers"
 
   ready_but_blocked = phase_ids - wave when blocked by file overlap, resource lock, or dependency frontier
@@ -94,6 +98,8 @@ for each wave in waves:
       run_started_at: state.started_at,
       recompose_count: decomposition.recompose_count,
       tier: tier.tier,
+      wave_index: wave_index,
+      timestamp: now_iso(),
       phases: phase_ids,
       wave,
       selected_mode: "parallel_rejected",
@@ -137,6 +143,8 @@ for each wave in waves:
       run_started_at: state.started_at,
       recompose_count: decomposition.recompose_count,
       tier: tier.tier,
+      wave_index: wave_index,
+      timestamp: now_iso(),
       phases: phase_ids,
       wave,
       selected_mode: "parallel",
@@ -154,6 +162,8 @@ for each wave in waves:
       run_started_at: state.started_at,
       recompose_count: decomposition.recompose_count,
       tier: tier.tier,
+      wave_index: wave_index,
+      timestamp: now_iso(),
       phases: phase_ids,
       wave,
       selected_mode: "parallel_failed",
@@ -211,8 +221,14 @@ unique per-run scope key is `state.started_at` (ISO timestamp set once at
   recompose. Without this key, a stale parallel event for tier N
   pre-recompose could satisfy the post-recompose tier N. Finalizer
   filters on this too.)
-- `timestamp` (event emission time; distinct from `run_started_at`)
 - `tier`
+- `wave_index` (per-tier wave counter, 0 for the skipped/sequential
+  branches and the zero-based wave index inside the `for each wave`
+  loop. Required so two same-tier `parallel_rejected` or
+  `parallel_failed` events do not collapse on the dedupe key.)
+- `timestamp` (ISO event-emission time; distinct from `run_started_at`.
+  Always set via `now_iso()` so two events emitted in the same
+  millisecond still differ when combined with `wave_index`.)
 - `phases`
 - `parallel_requested`
 - `selected_mode`: `skipped | sequential | parallel | parallel_rejected |
--- a/commands/mpl-run-execute.md
+++ b/commands/mpl-run-execute.md
@@ -43,6 +43,7 @@ phase_ids = tier.phases.filter(id => !is_completed(id))
 
 if phase_ids.length == 0:
   record_scheduler_event({
+    pipeline_id: state.pipeline_id,
     tier: tier.tier,
     phases: tier.phases,
     selected_mode: "skipped",
@@ -55,6 +56,7 @@ if phase_ids.length == 0:
 
 if tier.parallel != true or phase_ids.length == 1:
   record_scheduler_event({
+    pipeline_id: state.pipeline_id,
     tier: tier.tier,
     phases: phase_ids,
     selected_mode: "sequential",
@@ -79,6 +81,7 @@ for each wave in waves:
   ready_but_blocked = phase_ids - wave when blocked by file overlap, resource lock, or dependency frontier
   record ready_but_blocked_reason for each blocked phase in the tier reconciliation artifact
   record_scheduler_event({
+    pipeline_id: state.pipeline_id,
     tier: tier.tier,
     phases: phase_ids,
     wave,
@@ -134,8 +137,12 @@ run_cumulative_tests(phase_ids)
 
 `record_scheduler_event(...)` is mandatory telemetry. Append one JSON line to
 `.mpl/mpl/profile/phase-scheduler.jsonl` and mirror the latest 50 entries in
-`state.phase_scheduler_history`. Every row must include:
+`state.phase_scheduler_history`. The JSONL file is persistent across pipeline
+starts — `pipeline_id` on every row is what scopes events to the current run
+when the finalizer aggregates. Every row must include:
 
+- `pipeline_id` (= `state.pipeline_id` at write time; required so the
+  finalizer can filter out rows from prior runs in the same profile file)
 - `timestamp`
 - `tier`
 - `phases`
--- a/commands/mpl-run-execute.md
+++ b/commands/mpl-run-execute.md
@@ -42,9 +42,23 @@ For each tier in ascending `tier` order:
 phase_ids = tier.phases.filter(id => !is_completed(id))
 
 if phase_ids.length == 0:
+  record_scheduler_event({
+    tier: tier.tier,
+    phases: tier.phases,
+    selected_mode: "skipped",
+    parallel_requested: tier.parallel == true,
+    rejection_reasons: ["all_phases_already_completed"]
+  })
   continue
 
 if tier.parallel != true or phase_ids.length == 1:
+  record_scheduler_event({
+    tier: tier.tier,
+    phases: phase_ids,
+    selected_mode: "sequential",
+    parallel_requested: tier.parallel == true,
+    rejection_reasons: tier.parallel == true ? ["single_ready_phase"] : ["tier_parallel_false"]
+  })
   for each phase_id in phase_ids:
     execute_single_phase(phase_id)  // normal flow: 4.0.5 → 4.1 → 4.2 → 4.3 → 4.8
   continue
@@ -60,10 +74,22 @@ for each wave in waves:
 
   ready_but_blocked = phase_ids - wave when blocked by file overlap, resource lock, or dependency frontier
   record ready_but_blocked_reason for each blocked phase in the tier reconciliation artifact
+  record_scheduler_event({
+    tier: tier.tier,
+    phases: phase_ids,
+    wave,
+    selected_mode: wave.length > 1 ? "parallel" : "parallel_rejected",
+    parallel_requested: true,
+    worker_cap: max_phase_workers,
+    rejection_reasons_by_phase: ready_but_blocked_reason,
+    worktree_slots: wave.length > 1 ? planned_slot_ids : []
+  })
 
   // Multi-worktree pool. Reuse up to max_phase_workers isolated slots after
   // verifying each worktree has the current base branch commit reachable.
   worktree_pool = ensure_worktree_pool(size: min(max_phase_workers, wave.length))
+  append state.worktree_history entries for every slot creation/reuse:
+    { tier, phase_id, slot_id, worktree_path, base_ref, started_at, completed_at }
 
   results = parallel_map(wave, fn(phase_id, slot):
     seed = generate_phase_seed(phase_id, all_prior_summaries)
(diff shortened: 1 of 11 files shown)


KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new observability contract is internally inconsistent and is added to persisted state without the required migration path, so existing runs can still lose or emit unusable scheduler evidence.

Findings:
- [high] Scheduler event examples violate their own mandatory row schema (commands/mpl-run-execute.md:45-61)
  The skipped and sequential `record_scheduler_event` calls omit `worker_cap` and `worktree_slots`, but the same section declares those fields mandatory for every row. A run with completed tiers, non-parallel tiers, or single ready phases will append JSONL/state entries with a different shape than parallel rows. Any final-summary or post-run parser that relies on the documented mandatory fields can fail or silently misclassify why parallelism was not used.
  Recommendation: Include `worker_cap: max_phase_workers` and `worktree_slots: []` in skipped/sequential events, and add a test that validates the documented event shapes rather than only matching text.
- [medium] New persisted state field has no schema bump or migration (hooks/lib/mpl-state.mjs:130-133)
  `phase_scheduler_history` is added to `DEFAULT_STATE`, but `CURRENT_SCHEMA_VERSION` remains 6 and the migration registry still stops at v5-to-v6. Per the repo's migration policy, even optional fields with `[]` defaults require a schema bump and backfill. Existing `.mpl/state.json` files already at schema 6 will be returned unchanged by `readState`, so upgraded runs can have no `state.phase_scheduler_history` until a writer happens to replace it, undermining the promised state mirror and creating schema drift between fresh and existing projects.
  Recommendation: Bump `CURRENT_SCHEMA_VERSION`, add and register a v6-to-v7 migration that backfills `phase_scheduler_history: []`, update the schema-version comment/docs, and add a migration test for an existing v6 state.
- [medium] Final-summary requirement is not wired into finalize output (commands/mpl-run-execute.md:141-143)
  This section now requires the final run summary to explain why no `selected_mode:"parallel"` occurred when any execution tier requested parallelism, but the diff does not update the finalization prompt or run-summary schema that actually generate `.mpl/mpl/profile/run-summary.json`. Inference from the existing finalize command: Step 5.4 still documents only token/cache/phase/gate/totals fields, so this new MUST can be dropped at phase transition and the exact failure mode this change targets remains invisible in the summary.
  Recommendation: Update `commands/mpl-run-finalize.md` and `commands/schemas/run-summary.json` to read `phase-scheduler.jsonl` or `state.phase_scheduler_history` and emit a scheduler summary with the no-parallel explanation; cover it with a test.

Next steps:
- Fix the event schema first so every scheduler decision emits uniform telemetry.
- Add the schema migration before shipping to existing MPL workspaces.
- Wire scheduler history into the finalize summary path so the new MUST is enforceable.


— from codex

KyubumShin:
### Issues at 1eb71da

- **commands/mpl-run-execute.md:91-92** — The new parallel-pool writer appends entries to `state.worktree_history` with shape `{ tier, phase_id, slot_id, worktree_path, base_ref, started_at, completed_at }`, but the existing HIGH-risk isolation writer (commands/mpl-run-execute-context.md:383-393) writes the same array with an incompatible shape `{ phase_id, branch, path, risk_level, result, timestamp }`. The two writers now populate one array with non-overlapping field names (`path` vs `worktree_path`, no `branch`/`risk_level`/`result`/`timestamp` on the new shape, no `tier`/`slot_id`/`base_ref` on the old). Any reader (status report, archive, future hook) cannot rely on a uniform record shape. Suggested fix: either introduce a separate `state.worktree_pool_history` for the parallel-pool lifecycle, or extend the existing shape additively (add `tier`, `slot_id`, `base_ref`, `started_at`, `completed_at` while keeping `branch`/`path`/`result`/`timestamp` and a discriminator like `mode: "isolation" | "parallel_pool"`).
- **hooks/lib/mpl-state.mjs:130-133** — Adds `phase_scheduler_history: []` to `DEFAULT_STATE` without bumping `CURRENT_SCHEMA_VERSION` (still 6) or registering a migration under `hooks/lib/migrations/`. This contradicts the documented precedent set by `v5-to-v6.mjs` ("adding fields to DEFAULT_STATE without a migration would let legacy state.json read with `state.release.gate_results === undefined`"). Existing `.mpl/state.json` files written under schema v6 will be read back with `phase_scheduler_history === undefined`, so the prompt's "mirror the latest 50 entries in state.phase_scheduler_history" instruction silently turns into "create the field" on the first write — and any defensive read (`state.phase_scheduler_history.length`, `.slice`, `.map`) on legacy v6 state without `?? []` will throw. Suggested fix: bump `CURRENT_SCHEMA_VERSION` to 7 and add `hooks/lib/migrations/v6-to-v7.mjs` that backfills `phase_scheduler_history: []` on existing state, matching the additive-backfill pattern in v5-to-v6.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the claimed scheduler enforcement can silently pass when the telemetry it depends on is missing.

Findings:
- [high] Finalizer treats absent scheduler telemetry as no parallelism requested (commands/mpl-run-finalize.md:723-742)
  Step 5.4 derives both tiers_parallel_requested and tiers_parallel_executed solely from phase-scheduler events. If record_scheduler_event is skipped, phase-scheduler.jsonl is lost/corrupt, or state.phase_scheduler_history is empty, a decomposition with execution_tiers[].parallel:true collapses to tiers_parallel_requested=0 and no_parallel_explanation=null. That violates the new MUST because the final summary can hide the exact failure this change is meant to expose.
  Recommendation: Have finalize read .mpl/mpl/decomposition.yaml and compute the expected parallel-requested tiers from execution_tiers. Require one scheduler event per tier, and fail or emit a non-null no_parallel_explanation when any requested parallel tier has missing telemetry or no selected_mode:"parallel" event.

Next steps:
- Add a finalize test where decomposition has a parallel tier but phase-scheduler.jsonl/state history are empty; it should not produce tiers_parallel_requested=0 with a null explanation.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at ad039e4.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the scheduler summary can be satisfied by stale telemetry from an earlier run, so the new observability gate can silently lie.

Findings:
- [high] Scheduler aggregation is not scoped to the current run (commands/mpl-run-finalize.md:733-742)
  The finalizer reads every row from `.mpl/mpl/profile/phase-scheduler.jsonl` and then decides coverage/execution only by `event.tier`. That profile directory is preserved across pipeline starts (`hooks/lib/mpl-state.mjs` documents `.mpl/mpl/profile/` as persistent), while the new required scheduler event fields do not include `run_id` or `pipeline_id`. Because tier numbers are reused across decompositions, a previous run's `selected_mode:"parallel"` for tier 1 can make the current run report `tiers_parallel_executed` and avoid `tiers_with_missing_telemetry` even when the current scheduler never emitted telemetry. Impact: the exact no-parallel/missing-telemetry failure this change is meant to expose can finalize as clean.
  Recommendation: Make scheduler telemetry per-run: require `pipeline_id`/`run_id` in every scheduler event, filter JSONL/state events to the current state pipeline before computing scheduler fields, and add a regression test where stale profile rows from a prior run must not satisfy the current decomposition.

Next steps:
- Add run-scoping to the execute prompt/schema and finalizer aggregation contract.
- Add a stale `phase-scheduler.jsonl` regression test before shipping.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 0a6691b.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: scheduler telemetry scoping can still accept stale rows for repeat runs.

Findings:
- [high] Scheduler event scoping is not unique per run (commands/mpl-run-finalize.md:735-740)
  The finalizer filters persistent `.mpl/mpl/profile/phase-scheduler.jsonl` rows only by `pipeline_id`, but `pipeline_id` is not a unique run identifier; `initState` derives it from date plus feature slug, so restarting the same feature on the same day reuses the same value. A stale `selected_mode:"parallel"` row from the prior run can satisfy the current run's tier and hide missing telemetry, leaving `tiers_with_missing_telemetry` empty and `no_parallel_explanation` null even though the current scheduler never recorded the decision.
  Recommendation: Scope scheduler events with a genuinely unique run id, such as an ISO timestamp/random suffix recorded in state and in each event, or include `started_at`/event timestamp bounds in the finalizer filter. Add a regression test for two same-day same-slug runs sharing a persistent `phase-scheduler.jsonl`.

Next steps:
- Use a unique per-run key for scheduler telemetry and update executor/finalizer prompts plus tests to cover stale profile rows from repeated same-day runs.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 0765f5d.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: scheduler aggregation can misclassify multi-wave tiers as both executed and rejected, breaking the core observability guarantee this change is trying to add.

Findings:
- [medium] Parallel rejection count is inflated for tiers that actually execute in a later wave (commands/mpl-run-finalize.md:748-753)
  The executor records one event per wave and uses `selected_mode` based only on that wave's length. The finalizer then sets `tiers_parallel_executed` if any event for the tier is `parallel`, but computes `tiers_parallel_rejected = tiers_parallel_requested - tiers_parallel_executed`. In a tier split into multiple waves where one wave has length 1 and another has length >1, the tier has both a `parallel_rejected` and a `parallel` event. The summary will report the tier as executed and not rejected, dropping the single-wave rejection reason entirely from the rejected count. That makes the run-summary unable to explain partial parallelism within a requested tier, exactly the hard-to-diagnose case this telemetry is meant to surface.
  Recommendation: Aggregate at wave/event level or add a separate `tiers_with_partial_rejection`/`waves_parallel_rejected` metric; do not derive rejection as `requested - executed` once multiple scheduler events per tier are allowed.

Next steps:
- Adjust the finalizer scheduler model to preserve partial rejection evidence for tiers with mixed wave outcomes.
- Add a test with a parallel-requested tier that emits both `parallel_rejected` and `parallel` events for different waves.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 00027b2.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No ship: the observability path can produce false evidence for parallel execution and has an unhandled legacy/fallback finalize path.

Findings:
- [high] Parallel execution is logged before it is proven (commands/mpl-run-execute.md:85-96)
  The scheduler event is emitted before `ensure_worktree_pool(...)` and `parallel_map(...)` run. Finalize later treats any event with `selected_mode == "parallel"` as executed, so a pool setup failure, worker dispatch failure, or retry after a failed wave can leave a successful-looking row that satisfies the run summary even though parallel work never actually happened. That makes the new telemetry unreliable on the exact partial-failure path it is meant to expose.
  Recommendation: Separate planned decisions from execution evidence, or emit/update the `parallel` event only after worktree allocation and worker dispatch have succeeded. Add an outcome/error field and make finalize count only successful started/completed parallel waves.
- [medium] Finalize does not mirror the executor's missing-tier fallback (commands/mpl-run-finalize.md:729-746)
  The new aggregation reads `decomposition.execution_tiers` directly to build `expected_parallel_tiers` and `tiers_total`, while the executor still documents a legacy fallback when `execution_tiers` is missing or empty. A legacy/resumed run, or a run with the graph guard disabled, can execute via that fallback and then fail or misaggregate during finalize before writing `run-summary.json`, which the completion guard requires.
  Recommendation: Normalize scheduler tiers in finalize with the same fallback used by execute, or explicitly handle missing/empty `execution_tiers` as zero requested parallel tiers. Add a regression test for a decomposition with phases but no `execution_tiers`.

Next steps:
- Fix the scheduler event lifecycle so execution evidence cannot be written before execution starts/succeeds.
- Make finalize robust to missing or empty `execution_tiers` and cover that path with a test.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at b8afe5d.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the scheduler summary can still produce a clean-looking explanation state while requested parallel tiers were rejected or while the state fallback contains the real current-run telemetry.

Findings:
- [medium] Rejected parallel tiers can finalize with no explanation once any other tier ran in parallel (commands/mpl-run-finalize.md:797-807)
  The rollup computes rejected tiers separately, but the `no_parallel_explanation` requirement ignores `tiers_parallel_rejected` and `waves_parallel_rejected` unless the same tier also has a `parallel` event. A run with two `parallel:true` tiers where one emits `parallel` and another emits `sequential`/`parallel_rejected` will have `tiers_parallel_executed > 0`, no missing telemetry, no failed waves, and no same-tier partial rejection, so the summary is allowed to set `no_parallel_explanation` to null even though a requested tier never parallelized. That defeats the enforcement field this change adds for explaining lost parallelism.
  Recommendation: Trigger `no_parallel_explanation` when `tiers_parallel_executed < tiers_parallel_requested` or `tiers_parallel_rejected > 0`, and include the rejected tier ids, selected modes, and rejection reasons.
- [medium] State telemetry fallback is bypassed whenever the persistent JSONL parses (commands/mpl-run-finalize.md:745-756)
  The aggregation chooses `phase-scheduler.jsonl` before filtering and only falls back to `state.phase_scheduler_history` if the JSONL read is absent/failing. Because the profile directory is persistent, an old JSONL can parse successfully while current-run appends are missing or truncated; after filtering by `pipeline_id` and `run_started_at`, the current events can be empty and the state mirror containing the latest events is never consulted. This turns a degraded JSONL write into false missing telemetry or a wrong explanation.
  Recommendation: Read both sources, union and de-duplicate events before filtering, and only mark telemetry missing after neither JSONL nor `state.phase_scheduler_history` has an event for the expected tier/run.

Next steps:
- Add executable aggregation tests for mixed executed/rejected parallel tiers.
- Add a degraded-JSONL test where current events exist only in `state.phase_scheduler_history`.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 0bbbee9.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new finalize guard can be bypassed by a malformed or stale scheduler summary even when decomposition.yaml declares parallel tiers.

Findings:
- [high] Finalize guard trusts run-summary requested count instead of decomposition truth (hooks/mpl-require-finalize-artifacts.mjs:201-216)
  The guard first detects that decomposition.yaml has at least one parallel tier, but then decides whether an explanation is required from summary.scheduler.tiers_parallel_requested. If the finalizer emits a scheduler block with tiers_parallel_requested: 0, tiers_parallel_executed: 0, and no_parallel_explanation: null, explanationRequired is false and finalize_done is allowed. That is exactly the prompt-drift/missing-telemetry failure this hook is meant to block, and it lets a completed run hide rejected or unobserved parallelism.
  Recommendation: Use decomp.parallelTierCount as the authoritative requested count, and block if summary.scheduler.tiers_parallel_requested does not match it. Compute explanationRequired from the decomposition-derived count, not from the summary field.

Next steps:
- Add a regression test with decomposition parallel: true and run-summary.scheduler.tiers_parallel_requested set to 0/null with no explanation; it should block finalize.
- Update schedulerExplanationMissing to compare the summary scheduler counts against decomposition-derived tiers before accepting finalize_done=true.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at c8edba8.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new finalize guard can still certify scheduler observability from a self-reported summary without proof telemetry.

Findings:
- [high] Finalize guard trusts self-reported executed scheduler counts (hooks/mpl-require-finalize-artifacts.mjs:213-230)
  The guard derives only the requested parallel tier count from decomposition.yaml, then trusts run-summary.scheduler for tiers_parallel_executed, missing telemetry, partial rejection, and failed waves. It never recomputes those fields from phase-scheduler.jsonl or state.phase_scheduler_history, even though those event streams are the proof of execution. A prompt-drifted or stale run-summary can set tiers_parallel_requested to the correct decomposition count and tiers_parallel_executed to the same number, with empty missing/partial/failure arrays and null no_parallel_explanation, and this hook will allow finalize_done=true even if no current-run selected_mode:"parallel" event exists. That defeats the machine-enforced MUST this change is adding and lets false completion ship silently.
  Recommendation: Move scheduler aggregation into this hook or a shared helper: read and union .mpl/mpl/profile/phase-scheduler.jsonl with state.phase_scheduler_history, filter by pipeline_id, started_at, and recompose_count, recompute executed/missing/partial/failure counts, and block when the summary does not match the recomputed evidence.

Next steps:
- Add a regression test where decomposition declares a parallel tier, phase-scheduler telemetry is absent, and run-summary falsely reports tiers_parallel_executed == tiers_parallel_requested; the hook should block.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at cbe1b49.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new scheduler gate can miss valid parallel tiers and still permits drifted scheduler summaries to finalize.

Findings:
- [high] Valid execution_tiers YAML can bypass scheduler enforcement (hooks/lib/mpl-scheduler-aggregate.mjs:34-45)
  readDecompositionScope only recognizes tiers when the item starts exactly as `- tier: N` and `parallel:` appears later on its own line. Valid YAML forms already used in repo docs, such as inline maps (`- { tier: 4, phases: [...], parallel: true }`) or reordered mapping keys, parse as zero tiers or non-parallel tiers. Then schedulerExplanationMissing sees `tiers_parallel_requested === 0` and skips the run-summary/explanation checks, allowing a run that requested parallelism to finalize with no scheduler evidence.
  Recommendation: Parse `execution_tiers` with an order-insensitive YAML-capable path, or extend this parser to handle inline maps and arbitrary key order. Fail closed when `execution_tiers` exists but cannot be parsed, and add tests for inline/reordered tier mappings.
- [medium] Finalize guard allows scheduler summary drift (hooks/mpl-require-finalize-artifacts.mjs:195-212)
  The comment says run-summary must match computed raw evidence, but the guard only compares three scalar fields and array lengths. It does not compare `tiers_total`, `tiers_parallel_rejected`, `waves_parallel_rejected`, exact tier IDs in `tiers_with_missing_telemetry`/`tiers_with_partial_rejection`, or rejection reasons. A summary can underreport rejected waves or name the wrong missing tier IDs and still pass with any non-empty explanation, which defeats the observability guarantee this change is adding.
  Recommendation: Compare the full computed scheduler aggregate against `run-summary.scheduler`, including exact sorted arrays and wave rejection counts, or write/replace the scheduler block from the computed aggregate before finalization. Require the explanation to reference the computed problem tiers/reasons.

Next steps:
- Fix scheduler tier parsing and add bypass regression tests for inline and reordered YAML mappings.
- Tighten finalize validation so every scheduler summary field either exactly matches computed evidence or is generated from it.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 569702d.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

Do not ship: the new scheduler observability path can undercount rejected/failed waves and can accept meaningless explanations, which defeats the purpose of the enforcement.

Findings:
- [high] Scheduler events can collapse because the key relies on a timestamp the prompt examples omit (hooks/lib/mpl-scheduler-aggregate.mjs:134-135)
  `eventKey()` deduplicates events by pipeline/run/recompose/tier/timestamp/mode, but the executor's `record_scheduler_event` examples omit `timestamp` for every emitted mode. If the runner follows those examples, multiple `parallel_rejected` or `parallel_failed` waves in the same tier share an empty timestamp and collapse into one event. The computed aggregate can then underreport `waves_parallel_rejected` or `waves_parallel_failed`, and a matching run-summary can finalize with lost failure evidence.
  Recommendation: Add `timestamp` or, preferably, a required `event_id`/`wave_index` to every `record_scheduler_event` block, reject events missing it, and include that stable event identity in the dedupe key. Add a test with two same-tier rejected waves to prove both survive aggregation.
- [medium] The guard accepts any non-empty no-parallel explanation (hooks/mpl-require-finalize-artifacts.mjs:230-233)
  When parallelism is rejected, missing, or failed, the hook only checks that `scheduler.no_parallel_explanation` is a non-empty string. It never validates `summary.scheduler.rejection_reasons` against raw events and never checks that the explanation names the affected tiers, modes, or failure reasons required by the finalize contract. A summary with `no_parallel_explanation: "n/a"` can pass while preserving none of the diagnostic value needed to recover or audit scheduler failures.
  Recommendation: Return rejection/failure reasons and affected tier ids from `aggregateScheduler`, compare `summary.scheduler.rejection_reasons` to that computed set, and require the explanation to reference the computed affected tiers or fail closed.

Next steps:
- Patch the scheduler event identity contract and add aggregation tests for repeated same-tier waves.
- Strengthen finalize guard validation for explanation content and rejection reasons.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at bcd4fea.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the scheduler guard can be bypassed by valid YAML comments, and the finalize prompt can generate summaries that the hook later rejects.

Findings:
- [high] Valid YAML comments can disable scheduler enforcement (hooks/lib/mpl-scheduler-aggregate.mjs:45-65)
  `readDecompositionScope` only recognizes an `execution_tiers:` block when the key line is exactly followed by whitespace/newline, and it only treats `parallel` as true when the raw captured scalar exactly matches `true`. Valid YAML such as `execution_tiers: # scheduler` or `parallel: true # independent` is parsed as no tiers or `parallel=false` instead of failing closed. The finalizer then sees `tiers_parallel_requested === 0` and skips all scheduler checks, allowing a run that declared phase parallelism to finalize without telemetry or `no_parallel_explanation`.
  Recommendation: Use a YAML parser for `decomposition.yaml`, or at minimum strip YAML comments/normalize booleans correctly and fail closed when an `execution_tiers` item has a missing or invalid `parallel` field. Add tests for inline comments on both the block key and `parallel: true` scalars.
- [medium] Finalize prompt dedupes with a narrower key than the hook (commands/mpl-run-finalize.md:753-760)
  The finalizer instructions dedupe scheduler events by `(pipeline_id, run_started_at, tier, timestamp, selected_mode)`, but the emitted event contract and hook-side `eventKey` rely on `recompose_count` and `wave_index` to distinguish same-tier waves and pre/post-recompose rows. A prompt-following run-summary can collapse two rejected/failed waves with the same coarse timestamp, or keep a stale pre-recompose row and drop the current one before filtering. The hook recomputes with the wider key and will reject the generated summary, blocking a correct run until the summary is manually repaired.
  Recommendation: Make the prompt dedupe key match `hooks/lib/mpl-scheduler-aggregate.mjs:eventKey` exactly, including `recompose_count` and `wave_index`, and add a regression test that asserts those fields are present in the Step 5.4 dedupe key.

Next steps:
- Fix scheduler YAML parsing before relying on the finalize guard as a source of truth.
- Align the prompt aggregation contract with the hook aggregation key and add regression coverage.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 0879632.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No ship: the new scheduler guard can silently skip enforcement for a valid decomposition shape, allowing the exact no-parallel observability requirement to be bypassed.

Findings:
- [high] Valid inline execution_tiers bypasses scheduler enforcement (hooks/lib/mpl-scheduler-aggregate.mjs:54-55)
  readDecompositionScope only recognizes execution_tiers when the value begins on following indented lines. If decomposition.yaml uses a valid same-line YAML list such as execution_tiers: [{ tier: 1, phases: [phase-1, phase-2], parallel: true }], blockMatch is null and the function returns zero tiers. schedulerExplanationMissing then treats tiers_parallel_requested as 0 and skips all scheduler summary/explanation checks. I verified the existing phase-contract parser accepts that inline shape as valid, so this is a real bypass path for finalizing a run that requested parallelism without a scheduler block or no_parallel_explanation.
  Recommendation: Parse decomposition.yaml with a real YAML parser, or at minimum detect any present non-empty same-line execution_tiers value and either parse it or fail closed with __decomposition_unparseable__. Add a finalize-guard test for top-level inline execution_tiers with parallel: true.

Next steps:
- Fix the decomposition parser before shipping the guard.
- Add regression coverage for top-level inline execution_tiers and other unsupported non-empty execution_tiers forms.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at a950f2d.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No ship: the scheduler guard can both wedge valid finalization paths and fail open when its source of truth disappears.

Findings:
- [high] Sequential/skipped parallel tiers lose their rejection reasons (hooks/lib/mpl-scheduler-aggregate.mjs:352-363)
  The new executor contract emits `rejection_reasons` for `selected_mode: "skipped"` and `"sequential"` tier decisions, and the finalize prompt says the summary should union rejection reasons across all events. This aggregator only calls `collectRejectionReasons` for `parallel_rejected` and `parallel_failed`, so a `parallel: true` tier with one ready phase computes `rejection_reasons: []` instead of `single_ready_phase`. A correctly generated run-summary will be blocked by the finalize guard as a mismatch; matching the hook would omit the reason operators need to understand why parallelism was not used.
  Recommendation: Collect rejection/failure reasons from every current-run scheduler event that carries them, or at least from every non-`parallel` event for an expected parallel tier, and add tests for `sequential` and `skipped` events on `parallel: true` tiers.
- [medium] Missing decomposition.yaml bypasses scheduler enforcement (hooks/mpl-require-finalize-artifacts.mjs:183-184)
  `aggregateScheduler` returns `null` when `.mpl/mpl/decomposition.yaml` is missing or unreadable, and the finalize guard treats that as no scheduler problem. Because the scheduler denominator is explicitly supposed to come from decomposition.yaml, deleting or failing to read that artifact makes the guard skip the new scheduler checks entirely. The default finalize artifact list does not require decomposition.yaml, so a run can set `finalize_done=true` with no verifiable scheduler source of truth.
  Recommendation: Fail closed when decomposition.yaml is missing/unreadable during finalize, for example by returning `scheduler:decomposition_missing_or_unreadable`, or add decomposition.yaml to required completion evidence before running scheduler aggregation.

Next steps:
- Fix scheduler aggregation to include sequential/skipped rejection reasons and cover those modes in tests.
- Make the finalize guard fail closed when the decomposition source of truth cannot be read.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 6a8a7b9.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

Do not ship: the scheduler evidence gate can still be bypassed or made to reject summaries that follow the prompt contract.

Findings:
- [high] Missing scheduler contract disables finalize enforcement (hooks/lib/mpl-scheduler-aggregate.mjs:58-59)
  When `execution_tiers` is absent, `readDecompositionScope` returns an empty tier list instead of a parse/error state. Downstream, `schedulerExplanationMissing` exits when `tiers_parallel_requested === 0`, so a missing or truncated required scheduler contract silently disables the new finalize guard. Since `execution_tiers` is documented as required scheduler input, this lets a completed run finalize without proving whether phase parallelism was requested or rejected.
  Recommendation: Fail closed when `execution_tiers` is missing/empty in a present decomposition, and when `decomposition.yaml` is missing/unreadable during finalize, unless an explicit legacy/config override is in force.
- [medium] Parallel-event rejection reasons are dropped from the recomputed aggregate (hooks/lib/mpl-scheduler-aggregate.mjs:352-371)
  The executor prompt records `rejection_reasons_by_phase` even on `selected_mode: "parallel"` events, and finalize Step 5.4 says `rejection_reasons` is the union across all events. The hook only calls `collectRejectionReasons` for `parallel_rejected`, `parallel_failed`, `sequential`, and `skipped`, so a correctly generated summary that includes reasons from a successful parallel wave will be blocked as a mismatch; omitting them to satisfy the hook under-reports scheduler blocking evidence.
  Recommendation: Collect rejection reasons for every filtered scheduler event, or at minimum include `selected_mode === "parallel"`, and add a regression test with a parallel event carrying `rejection_reasons_by_phase`.

Next steps:
- Make the scheduler aggregate fail closed on missing required decomposition scheduler data.
- Align rejection reason collection with the finalize prompt’s all-events union rule.


— from codex

KyubumShin:
### Issues at 36a91d4

- **hooks/lib/mpl-scheduler-aggregate.mjs:352-373** — `collectRejectionReasons` is invoked for `parallel_rejected`, `parallel_failed`, `sequential`, and `skipped` events but NOT for `parallel` events, while `commands/mpl-run-finalize.md:820-822` instructs the finalize prompt to compute `rejection_reasons` as the "union of event.rejection_reasons and the values of event.rejection_reasons_by_phase across all events (deduplicated)". The executor pseudocode at `commands/mpl-run-execute.md:141-155` explicitly writes `rejection_reasons_by_phase: ready_but_blocked_reason` on the successful `parallel` event itself, so any tier that splits into multiple conflict-free waves (the partial-parallelism scenario this telemetry was designed for) emits a `parallel` event whose `rejection_reasons_by_phase` is non-empty (e.g. `{"phase-3": ["file_overlap"]}` when phase-3 is deferred to a later wave). A prompt-faithful summary will include `"file_overlap"` in `rejection_reasons`; the hook will compute `[]`; finalize will then block with `scheduler:rejection_reasons_mismatch:computed=[],summary=[file_overlap]` even though the run executed correctly and the summary follows the documented contract. This is the same class of mismatch the r15 [high] fix addressed for sequential/skipped events — the fix is incomplete because it left `selected_mode === 'parallel'` events out of the union. Suggested fix: either drop the `selected_mode` guard entirely and call `collectRejectionReasons(e)` for every event in the loop (so the hook matches the prompt's "across all events" rule), or narrow the prompt's `rejection_reasons` definition to "across non-parallel events" and align the executor to only emit `rejection_reasons_by_phase` on rejection/failure events. The first option keeps the wave-split partial-parallelism signal observable in `rejection_reasons` and matches the existing prompt wording.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No ship: the scheduler guard can still accept hollow explanations and incomplete tier telemetry, so the new observability guarantee is not enforceable yet.

Findings:
- [high] no_parallel_explanation can be a meaningless placeholder (hooks/mpl-require-finalize-artifacts.mjs:248-264)
  The guard only checks that no_parallel_explanation is non-empty and contains the affected tier numbers. It does not require the selected mode, rejection reason, or failure reason that the finalize contract says the explanation must provide. A rejected tier with computed rejection_reasons=['file_overlap'] would pass with no_parallel_explanation='tier 1', leaving operators with no actual reason while finalize_done is allowed.
  Recommendation: Validate the explanation against computed evidence: require the affected selected mode and at least one computed rejection/failure reason, or replace this free-text gate with a structured per-tier explanation object that is compared to the aggregate.
- [high] One parallel event satisfies an entire multi-wave tier (hooks/lib/mpl-scheduler-aggregate.mjs:370-375)
  aggregateScheduler marks a tier executed when any scoped event for that tier has selected_mode='parallel', and marks telemetry present when any event for the tier exists. Because the decomposition parser drops tier phase membership and there is no wave/phase coverage check, a multi-wave tier can emit one successful parallel event, lose the later wave telemetry, and still compute tiers_parallel_executed=1 with no missing telemetry or required explanation. This inference follows from the executor contract emitting events per wave via wave_index.
  Recommendation: Retain phases per execution_tier and verify coverage from event.wave/selected_mode records, or add an explicit expected wave count. Do not count a tier fully executed until all expected phases/waves are accounted for by successful parallel events or explained rejection/failure events.

Next steps:
- Add adversarial tests for placeholder explanations such as 'tier 1' and for partial multi-wave telemetry loss.
- Tighten the aggregate/schema comparison after the coverage model is explicit.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 637db5e.

— from claude

PR #216, opened 2026-05-27
Add mpl-hook-trace for hook chain visibility (#207)

### LINKED ISSUE #207: [Exp22] Add mpl-hook-trace for hook chain visibility
Source: ygg-exp22 V1 execution analysis, R10

Exp22 spent about 10 minutes manually enumerating which hooks could block writes to `.mpl/mpl/decomposition.yaml`. The report proposes a hook-chain visibility tool.

Current repo state:

- Hook blocking has improved through `blocked_hook` envelopes and `/mpl:mpl-recover`.
- There is no dedicated `mpl-hook-trace <file>` style tool or command that lists which hooks target a path/event and why.

Scope:

- Add a local trace command, script, or skill entry that accepts a file path and reports the relevant MPL hook chain.
- Include at least:
  - PreToolUse hooks that can block writes to the path
  - PostToolUse hooks that can block continuation after relevant tool events
  - hook script path
  - event/matcher, when available
  - short purpose summary
  - current block status when state/config/context makes that check computable
- Make `.mpl/mpl/decomposition.yaml` the first-class test case and include:
  - goal-trace/hash drift check
  - baseline guard
  - artifact schema validation
  - discovery scanner
  - chain assignment
  - phase controller
  - require-test-agent continuation gate
- Add tests or fixtures that ensure the trace output stays aligned with `hooks/hooks.json`.

Acceptance criteria:

- `mpl-hook-trace .mpl/mpl/decomposition.yaml` or an equivalent supported invocation lists the hook chain without manual grep.
- Output distinguishes "registered hook" from "currently blocking".
- The tool is documented in the recovery/debugging path.

KyubumShin:
## What

- `hooks/lib/mpl-hook-trace.mjs`: traces the MPL hook chain for a given path, listing PreToolUse / PostToolUse hooks that can block writes, their event/matcher, purpose summary, and current block status when state/config/context allows it. Loads the registered chain from `hooks/hooks.json` so the output stays aligned with the real registry. Distinguishes "registered hook" from "currently blocking" by reading `state.blocked_hook`.
- `hooks/__tests__/mpl-hook-trace.test.mjs`: pins the `.mpl/mpl/decomposition.yaml` trace (canonical artifact for the feature) and asserts that an active `blocked_hook` envelope flips the `currently_blocking` flag for the relevant entry.
- `README.md`: adds the invocation line `node hooks/lib/mpl-hook-trace.mjs .mpl/mpl/decomposition.yaml` alongside the existing recovery commands.

## Why

Exp22 R10 spent ~10 minutes manually enumerating which hooks could block writes to `.mpl/mpl/decomposition.yaml`. With the tracer, the same investigation completes in seconds and stays aligned with the registry as it evolves.

## Design

N/A — small additive tool. Motivation, scope, and acceptance criteria live in issue #207 (ygg-exp22 V1 execution analysis, R10).

## Agent Review Status

- [ ] from claude
- [ ] from codex

Closes #207.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -487,6 +487,9 @@ mpl small add retry logic                     # → forces Standard
 
 # Hook-block recovery after resume reports blocked_hook
 /mpl:mpl-recover
+
+# Hook-chain visibility for a canonical artifact
+node hooks/lib/mpl-hook-trace.mjs .mpl/mpl/decomposition.yaml
 ```
 
 Codex also discovers the same MPL skills from the plugin. Use `mpl ...` in a Codex session, or invoke the installed MPL skill by name from the skills menu.
--- a/hooks/__tests__/mpl-hook-trace.test.mjs
+++ b/hooks/__tests__/mpl-hook-trace.test.mjs
@@ -68,6 +68,35 @@ describe('mpl-hook-trace', () => {
     }
   });
 
+  it('surfaces incomplete blocked_hook envelopes as invalid instead of currently_blocking', () => {
+    // Codex r2 on PR #216: a zombie state with blocked_by_hook set but
+    // empty/missing blocked_artifact must not be reported as actively
+    // blocking — that would point recovery at the wrong artifact and
+    // hide the real state-invariant failure.
+    const tmp = mkdtempSync(join(tmpdir(), 'mpl-hook-trace-invalid-'));
+    try {
+      mkdirSync(join(tmp, '.mpl'), { recursive: true });
+      writeFileSync(join(tmp, '.mpl', 'state.json'), JSON.stringify({
+        schema_version: CURRENT_SCHEMA_VERSION,
+        current_phase: 'mpl-decompose',
+        session_status: 'blocked_hook',
+        blocked_by_hook: 'mpl-require-goal-trace',
+        // blocked_artifact intentionally absent
+      }));
+      const trace = traceHookChain({
+        targetPath: '.mpl/mpl/decomposition.yaml',
+        cwd: tmp,
+      });
+      const row = trace.hooks.find((h) => h.hook_id === 'mpl-require-goal-trace');
+      assert.equal(row.status, 'invalid_blocked_envelope');
+      assert.match(formatHookTrace(trace), /INVALID_BLOCKED_ENVELOPE/);
+      assert.doesNotMatch(formatHookTrace(trace), /\[BLOCKING\]/,
+        'must not surface an incomplete envelope as actively blocking');
+    } finally {
+      rmSync(tmp, { recursive: true, force: true });
+    }
+  });
+
   it('marks the active blocked_hook as currently_blocking', () => {
     const tmp = mkdtempSync(join(tmpdir(), 'mpl-hook-trace-'));
     try {
--- a/hooks/__tests__/mpl-hook-trace.test.mjs
+++ b/hooks/__tests__/mpl-hook-trace.test.mjs
@@ -1,0 +1,54 @@
+import { describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+import { mkdtempSync, mkdirSync, rmSync, writeFileSync } from 'fs';
+import { join } from 'path';
+import { tmpdir } from 'os';
+
+import { formatHookTrace, traceHookChain } from '../lib/mpl-hook-trace.mjs';
+import { CURRENT_SCHEMA_VERSION } from '../lib/mpl-state.mjs';
+
+describe('mpl-hook-trace', () => {
+  it('lists the decomposition hook chain without manual grep', () => {
+    const trace = traceHookChain({
+      targetPath: '.mpl/mpl/decomposition.yaml',
+      cwd: process.cwd(),
+    });
+    const ids = trace.hooks.map((h) => h.hook_id);
+    for (const expected of [
+      'mpl-require-goal-trace',
+      'mpl-baseline-guard',
+      'mpl-artifact-schema',
+      'mpl-discovery-scanner',
+      'mpl-require-chain-assignment',
+      'mpl-phase-controller',
+      'mpl-require-test-agent',
+    ]) {
+      assert.ok(ids.includes(expected), `missing ${expected}`);
+    }
+    assert.match(formatHookTrace(trace), /MPL Hook Trace/);
+    assert.match(formatHookTrace(trace), /registered/);
+  });
+
+  it('marks the active blocked_hook as currently_blocking', () => {
+    const tmp = mkdtempSync(join(tmpdir(), 'mpl-hook-trace-'));
+    try {
+      mkdirSync(join(tmp, '.mpl'), { recursive: true });
+      writeFileSync(join(tmp, '.mpl', 'state.json'), JSON.stringify({
+        schema_version: CURRENT_SCHEMA_VERSION,
+        current_phase: 'mpl-decompose',
+        session_status: 'blocked_hook',
+        blocked_by_hook: 'mpl-require-goal-trace',
+        blocked_artifact: '.mpl/mpl/decomposition.yaml',
+      }));
+      const trace = traceHookChain({
+        targetPath: '.mpl/mpl/decomposition.yaml',
+        cwd: tmp,
+      });
+      const row = trace.hooks.find((h) => h.hook_id === 'mpl-require-goal-trace');
+      assert.equal(row.status, 'currently_blocking');
+      assert.match(formatHookTrace(trace), /BLOCKING/);
+    } finally {
+      rmSync(tmp, { recursive: true, force: true });
+    }
+  });
+});
--- a/hooks/__tests__/mpl-hook-trace.test.mjs
+++ b/hooks/__tests__/mpl-hook-trace.test.mjs
@@ -129,6 +129,36 @@ describe('mpl-hook-trace', () => {
     }
   });
 
+  it('surfaces a stale/invalid block on a non-decomposition path as invalid_blocked_envelope (does not silently filter)', () => {
+    // Codex r5 on PR #216: a stale envelope (missing retry_context etc.)
+    // pointing at state.test_agent_dispatched.<phase> would otherwise be
+    // filtered out by the category check and disappear from the trace.
+    // The offending hook id is still meaningful to the operator and must
+    // appear, just with an invalid_blocked_envelope status.
+    const tmp = mkdtempSync(join(tmpdir(), 'mpl-hook-trace-stale-task-'));
+    try {
+      mkdirSync(join(tmp, '.mpl'), { recursive: true });
+      writeFileSync(join(tmp, '.mpl', 'state.json'), JSON.stringify({
+        schema_version: CURRENT_SCHEMA_VERSION,
+        current_phase: 'phase2-sprint',
+        session_status: 'blocked_hook',
+        blocked_by_hook: 'mpl-require-test-agent',
+        blocked_artifact: 'state.test_agent_dispatched.phase-1',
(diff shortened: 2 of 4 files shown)


KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new trace diagnostic can mutate MPL run state just by being executed.

Findings:
- [high] Hook trace uses a mutating state reader (hooks/lib/mpl-hook-trace.mjs:142)
  `traceHookChain` calls `readState(cwd)` whenever `.mpl/state.json` exists. In this repo, `readState` persists migrations for old or missing `schema_version`, and the v1-to-v2 migration can archive and remove `.mpl/mpl/state.json`. That makes a visibility/recovery command capable of changing active pipeline state before the operator has diagnosed the blocked hook, which is a rollback/version-skew and state-forensics hazard. This is not just theoretical: the reviewed workspace state file lacks `schema_version`, so this path is plausible for existing MPL worktrees.
  Recommendation: Do not call the mutating `readState` from this diagnostic path. Add a read-only state loader or a `readState(..., { migrate: false })` mode that parses `.mpl/state.json` without migration/archive/removal, and add a regression test that tracing a legacy state leaves `.mpl/state.json` and `.mpl/mpl/state.json` byte-for-byte unchanged.

Next steps:
- Make hook tracing strictly read-only before shipping.
- Add a no-mutation test for legacy state/schema-version cases.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 0899e2a.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new recovery diagnostic can present invalid blocked-hook state as an authoritative active block.

Findings:
- [medium] Invalid blocked_hook envelope is reported as actively blocking (hooks/lib/mpl-hook-trace.mjs:137-145)
  `blockStatusFor` returns `currently_blocking` when the state has `session_status:'blocked_hook'` and matching `blocked_by_hook` but no `blocked_artifact` (`!artifact`). That companion field is required for an actionable hook block, so a stale or partially cleared zombie state will print `BLOCKING` for the requested target instead of surfacing that the block envelope is corrupt. In recovery, that can send the operator toward editing/retrying the wrong artifact and hide the real state-invariant failure.
  Recommendation: Validate the blocked-hook envelope before assigning `currently_blocking`; require a non-empty `blocked_artifact` and normalized target match, and emit an explicit invalid/unknown blocked state when required companion fields are missing. Add regression coverage for missing `blocked_artifact` and other missing companion fields.

Next steps:
- Tighten blocked-hook state validation in the tracer before rendering `BLOCKING`.
- Add tests for incomplete blocked_hook envelopes so recovery diagnostics fail visibly instead of looking normal.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at a4cf4a0.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new tracer can still label corrupt blocked_hook state as actively blocking, which undermines recovery diagnostics.

Findings:
- [medium] Corrupt blocked_hook envelopes still show as BLOCKING (hooks/lib/mpl-hook-trace.mjs:146-152)
  `blockStatusFor` only requires `blocked_artifact` to be non-empty and path-matched before returning `currently_blocking`. Existing state invariants define an actionable blocked hook as also requiring `blocked_phase`, `block_code`, `block_reason`, `resume_instruction`, `blocked_at`, and object `retry_context`; a state missing those fields is explicitly stale/corrupt. Under that real failure mode, the tracer prints `[BLOCKING]` instead of surfacing an invalid envelope, sending operators toward normal hook recovery and hiding the state-invariant failure.
  Recommendation: Validate the same companion fields required by the blocked_hook invariant before returning `currently_blocking`, or reuse/export a shared blocked-hook envelope validator; otherwise return `invalid_blocked_envelope` with the missing fields.

Next steps:
- Add a test where `session_status`, `blocked_by_hook`, and `blocked_artifact` are present but `block_code`/`retry_context` or other required companion fields are missing, and assert the tracer does not emit `[BLOCKING]`.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at e9e2cec.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the trace can falsely report no active block for a first-class blocked_hook path.

Findings:
- [medium] Active test-agent blocks are filtered out before status detection (hooks/lib/mpl-hook-trace.mjs:122-136)
  `shouldIncludeHook` only includes `Task|Agent` PostToolUse hooks when the target is categorized as `decomposition`. A valid `mpl-require-test-agent` block records `blocked_artifact` as `state.test_agent_dispatched.<phase>`, which `pathCategory` treats as a generic file, so the `mpl-require-test-agent` registration is skipped and `blockStatusFor` is never called for the blocking hook. Reproduced with an inline valid blocked_hook state: tracing `state.test_agent_dispatched.phase-1` produced no `BLOCKING` row and omitted `mpl-require-test-agent`. This defeats the recovery/diagnostic purpose for AD-0007 blocks and can send operators looking at the wrong hooks.
  Recommendation: When state is `session_status='blocked_hook'` and the target matches `blocked_artifact`, include the `blocked_by_hook` registration regardless of the category/matcher filter, or add an explicit pseudo-artifact category for `state.test_agent_dispatched.*` that includes the Task/Agent hook chain. Add a regression test for tracing `state.test_agent_dispatched.phase-1` with a valid `mpl-require-test-agent` block.

Next steps:
- Fix the inclusion logic and add the missing regression before shipping.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at d3caf57.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No ship: the new diagnostic still hides a stale active hook block in the exact non-file path case it was added to make recoverable.

Findings:
- [medium] Invalid active Task/Agent hook blocks are filtered out instead of reported (hooks/lib/mpl-hook-trace.mjs:183-204)
  `activeBlockHookId` is only set when `missingBlockedHookFields(activeState).length === 0`. If `session_status='blocked_hook'` points at a non-decomposition artifact such as `state.test_agent_dispatched.phase-1` but the envelope is stale, e.g. missing `retry_context`, line 204 falls back to the normal category/matcher filter. For category `file`, the `Task|Agent` `mpl-require-test-agent` hook is then omitted, so the trace shows no `INVALID_BLOCKED_ENVELOPE` for the hook that is actually keeping the run blocked. I reproduced this by calling `traceHookChain` with `blocked_by_hook: 'mpl-require-test-agent'`, matching `blocked_artifact`, and no `retry_context`; the formatted trace omitted `mpl-require-test-agent`, while adding `retry_context` made it show as `BLOCKING`. This leaves operators blind to the stale envelope and pushes recovery toward the wrong place.
  Recommendation: Surface the recorded `blocked_by_hook` whenever `session_status === 'blocked_hook'` and the target matches `blocked_artifact`, even if required companion fields are missing; let `blockStatusFor` mark that row as `invalid_blocked_envelope`. Add a regression test for `mpl-require-test-agent` with a matching artifact and missing `retry_context`/`block_code`.

Next steps:
- I could not run the new test file cleanly in this read-only sandbox because its temp-directory setup fails with `EPERM`; the logic above was verified with an in-memory `traceHookChain` call.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 73d4ee5.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the diagnostic can misreport incompatible or corrupt run state as an actionable active hook block.

Findings:
- [medium] Raw state read bypasses fail-closed schema validation (hooks/lib/mpl-hook-trace.mjs:14-20)
  `mpl-hook-trace` avoids `readState()` to stay read-only, but the replacement raw parser accepts any parseable JSON and only returns null on parse failure. That drops the existing `readState()` protections for malformed state shape and `schema_version > CURRENT_SCHEMA_VERSION`; a downgraded plugin can therefore classify a future/incompatible `blocked_hook` envelope as `currently_blocking` and send recovery toward artifact edits instead of surfacing the required upgrade/restore path. This is a version-skew recovery hazard, inferred from the raw parser here plus `blockStatusFor` only checking companion fields before emitting BLOCKING.
  Recommendation: Keep the byte-for-byte raw read, but validate the parsed object before using it: require an object with the minimal state shape, fail closed on `schema_version > CURRENT_SCHEMA_VERSION`, and surface an explicit `unsupported_schema`/`unreadable_state` status instead of treating it as no block or a valid block. Add regression tests for future schema and corrupt JSON.

Next steps:
- Add read-only raw-state validation that mirrors `readState()` fail-closed checks without running migrations.
- Add hook-trace tests for unsupported `schema_version` and malformed `.mpl/state.json`.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 1d905bc.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the trace can hide the active blocker under hook registry/version skew.

Findings:
- [medium] Active blocked_hook is omitted when the hook is no longer registered (hooks/lib/mpl-hook-trace.mjs:226-232)
  The force-include path only runs while iterating hooks from hooks.json, so it can include the active blocker only if that hook id is still present in the current registry. If a run is resumed after an upgrade/downgrade or a hooks.json skew where state.blocked_by_hook names a removed/renamed hook, the trace emits no BLOCKING or INVALID_BLOCKED_ENVELOPE row even though the state is blocked. That makes the diagnostic unsafe for exactly the recovery/version-skew case it is meant to clarify.
  Recommendation: After building rows, if activeBlockHookId matches the target and no row has that hook_id, append a synthetic row such as event='state', matcher='blocked_by_hook', command=null, purpose='active blocker not registered in hooks.json', with status from blockStatusFor(), and add a regression test for an unknown blocked_by_hook.

Next steps:
- Patch the synthetic active-blocker row and add a regression test covering removed/renamed hook ids.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at e73411d.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the diagnostic can still hide the exact stale blocked_hook state it was added to expose.

Findings:
- [medium] Active blocked_hook can disappear when the invalid envelope is missing blocked_artifact (hooks/lib/mpl-hook-trace.mjs:216-232)
  The force-include path depends on activeBlockMatchesTarget, but that is false when blocked_artifact is absent or blank. blocked_artifact is one of the required fields for the blocked_hook envelope, so this misses a real stale-state case: session_status remains blocked_hook, blocked_by_hook points at mpl-require-test-agent, blocked_artifact is missing, and tracing state.test_agent_dispatched.<phase> filters the Task|Agent hook out as a generic file target. The trace then shows no INVALID_BLOCKED_ENVELOPE row even though the run is still paused by blocked_hook, pushing recovery away from the state-invariant failure.
  Recommendation: Detect an invalid active blocked_hook envelope before target matching. If blocked_by_hook is present and missingBlockedHookFields(activeState) is non-empty, emit the active hook row with invalid_blocked_envelope even when blocked_artifact is missing, and add a synthetic row when the hook id is not registered.

Next steps:
- Add a regression test for a blocked_hook missing blocked_artifact on a Task|Agent blocker and a non-decomposition target.
- Move envelope validation ahead of activeBlockMatchesTarget or make the force-include condition include invalid active envelopes.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at f147d93.

— from claude

Issue #208, opened 2026-05-27
[Exp22] Harden mpl-test-agent schema compliance recovery


KyubumShin:
Source: ygg-exp22 V1 execution analysis, R7

Exp22 accumulated 14 `INVALID` mpl-test-agent verdicts because the test-agent often returned a long natural-language report instead of the required JSON evidence block. Some of those runs performed useful testing and found bugs, but the hook evidence remained invalid.

Current repo state:

- `agents/mpl-test-agent.md` already declares that final output must be a valid JSON block wrapped in json fences.
- `hooks/lib/mpl-test-agent-evidence.mjs` already parses fenced JSON or the first JSON object candidate and validates the evidence contract strictly.
- `hooks/mpl-require-test-agent.mjs` now blocks through a resumable `blocked_hook` envelope when required evidence is missing or invalid.

The remaining gap is recovery UX and repeated schema noncompliance, not basic parser support.

Scope:

- Strengthen the test-agent final-output rule:
  - the final assistant message must start with a `json` fence
  - the final assistant message must contain no prose before or after the JSON block
  - natural-language summaries are allowed only inside JSON fields
- When invalid test-agent evidence is recorded, preserve enough diagnostics for a direct retry:
  - invalid reason
  - response length
  - short response preview
  - phase id
  - required schema reminder
  - exact re-dispatch prompt for mpl-test-agent
- Surface that retry prompt in `/mpl:mpl-resume` or `/mpl:mpl-recover` output when the current blocker is `missing_or_invalid_test_agent_evidence`.
- Keep strict evidence semantics. Do not infer PASS from natural-language reports.
- Consider model/profile escalation only after repeated INVALID responses for the same phase, and record that escalation as an explicit policy choice.

Acceptance criteria:

- A malformed natural-language test-agent response produces a clear blocked-hook resume path, not a dead end.
- The retry prompt is copy-ready and includes the exact schema/fence requirement.
- Tests cover at least:
  - prose-only test-agent output
  - JSON object without required fields
  - valid fenced JSON
  - repeated INVALID response preserving retry diagnostics

KyubumShin:
Closed by PR #218 (commit c6178c1).

PR #218, opened 2026-05-27
Detect subagent anomalies + harden test-agent recovery (#206, #208)

### LINKED ISSUE #206: [Exp22] Detect zero-token or empty subagent returns
Source: ygg-exp22 V1 execution analysis, R8

Exp22 observed four subagent completions with apparent zero-token or empty final returns. Three of them used many tools and wrote files, but produced no final assistant message:

- phase-19 phase-runner: 58 tools, 0 tokens, about 55m
- phase-20 test-agent: 32 tools, 0 tokens, about 35m
- phase-21 phase-runner: 40 tools, 0 tokens, about 28m

Current repo state:

- `hooks/mpl-validate-output.mjs` estimates token count from response text length for output validation/profiling.
- There does not appear to be a dedicated anomaly detector for an empty subagent response after substantial tool activity.

Scope:

- Detect subagent completion anomalies when the hook payload exposes enough data:
  - empty or missing final response
  - zero output tokens, if usage metadata is available
  - substantial tool activity, duration, or file changes
- Record a structured anomaly entry in state/profile output:
  - phase id, agent type, response length, output tokens if available, tool count if available, duration if available, affected files if available
  - anomaly type: `empty_response`, `zero_token_after_tools`, or `agent_init_failure`
- For phase-runner anomalies, recommend on-disk artifact verification or re-dispatch rather than silently treating the phase as cleanly completed.
- For mpl-test-agent anomalies, mark evidence as invalid with a specific reason such as `empty_response_anomaly`, so `/mpl:mpl-resume` can suggest the test-agent retry path.
- Add tests with synthetic hook payloads for:
  - empty response with no tools
  - empty response with many tools
  - normal non-empty response

Acceptance criteria:

- Empty subagent final messages are visible in MPL artifacts and stop/resume guidance.
- A future exp report does not need screen/log archaeology to identify zero-token return anomalies.
- The detector is best-effort and does not depend on provider-specific metadata being present.

KyubumShin:
## What

Bundles two coupled Exp22 issues — they share the gate-recorder hook payload, the test-agent evidence parser, and recovery-skill copy.

**#206 — Subagent return anomaly detection**
- `hooks/lib/mpl-subagent-anomaly.mjs` (new): classifies each Task|Agent completion as `empty_response` / `zero_token_after_tools` / `agent_init_failure` based on response length, usage tokens, tool count, duration. Persists structured entries into `state.subagent_return_anomalies` (ring-capped). Returns a formatted operator notice.
- `hooks/mpl-gate-recorder.mjs`: wires the detector + threads anomaly into the evidence parser + propagates a `systemMessage` on continue responses.
- `hooks/__tests__/mpl-subagent-anomaly.test.mjs` (new): detect / append / format across the three anomaly types and a normal run.

**#208 — Test-agent schema compliance recovery**
- `agents/mpl-test-agent.md`: final message MUST start with a `json` fence; natural language only inside JSON fields.
- `hooks/lib/mpl-test-agent-evidence.mjs`: accepts an `anomaly` arg, preserves a truncated `response_preview` (with `[truncated]` marker), and emits `invalid_reason: 'empty_response_anomaly'` for detector-flagged empty returns.
- `hooks/mpl-require-test-agent.mjs`: missing/invalid evidence now produces a resumable `blocked_hook` envelope whose `retry_context` includes the invalid reason, preview, schema reminder, and re-dispatch prompt.
- `skills/mpl-recover/SKILL.md`: documents the surfaced retry path.

Coverage in `hooks/__tests__/mpl-gate-recorder.test.mjs` and `hooks/__tests__/mpl-require-test-agent.test.mjs` (prose-only output, anomaly-flagged empty returns, response_preview truncation, re-dispatch prompt, resumable envelope).

## Why

Exp22 R7/R8 surfaced two failure modes that the current pipeline mishandled:

- Four subagent completions returned 0-token output after substantial tool activity. The hook payload had enough data to flag them, but no detector recorded them and the runs looked cleanly completed.
- 14 INVALID mpl-test-agent verdicts accumulated from prose-only responses. Some runs found real bugs but the evidence-parser dead-ended and `/mpl:mpl-resume` had no concrete retry path.

Bundling them in one PR avoids two passes over `mpl-gate-recorder.mjs` and keeps the evidence-vs-anomaly handoff atomic.

## Design

N/A — additive detector + parser strengthening. Motivation, scope, and acceptance criteria live in issues #206 and #208 (ygg-exp22 V1 execution analysis, R7+R8).

## Agent Review Status

- [ ] from claude
- [ ] from codex

Closes #206, #208.

---

### DIFF
--- a/agents/mpl-test-agent.md
+++ b/agents/mpl-test-agent.md
@@ -84,7 +84,12 @@ disallowedTools: Task
   </Execution_Flow>
 
   <Output_Schema>
-    Your final output MUST be a valid JSON block wrapped in ```json fences.
+    FINAL OUTPUT RULE:
+    - Your final assistant message MUST start with a ```json fence.
+    - Your final assistant message MUST end with the closing ``` fence.
+    - Do NOT put prose, headings, bullets, or commentary before or after the JSON block.
+    - Put any human-readable summary inside JSON fields only.
+    - A natural-language final report is INVALID even when tests passed.
 
     ```json
     {
--- a/hooks/__tests__/mpl-gate-recorder.test.mjs
+++ b/hooks/__tests__/mpl-gate-recorder.test.mjs
@@ -392,6 +392,50 @@ describe('mpl-gate-recorder test-agent evidence', () => {
     assert.equal(ev.valid_json, false);
     assert.equal(ev.invalid_reason, 'missing_json_block');
   });
+
+  it('phase-runner anomaly does not advance sprint_status.completed_todos (codex r3)', () => {
+    // Codex r3 on PR #218: when the phase-runner returns anomalous output
+    // (empty / zero tokens after substantial tool work), the hook must not
+    // count a possibly-stale state-summary.md as a completed phase.
+    const tmp = mkdtempSync(join(tmpdir(), 'mpl-gate-recorder-phase-runner-anomaly-'));
+    try {
+      seedState(tmp, { sprint_status: { completed_todos: 0, in_progress_todos: 0, failed_todos: 0, total_todos: 1 } });
+      // Plant a stale on-disk state-summary.md so countCompletedPhases would
+      // otherwise increment to 1.
+      mkdirSync(join(tmp, '.mpl', 'mpl', 'phases', 'phase-1'), { recursive: true });
+      writeFileSync(
+        join(tmp, '.mpl', 'mpl', 'phases', 'phase-1', 'state-summary.md'),
+        '# phase-1 state summary\nstatus: COMPLETED\n'
+      );
+      const input = {
+        cwd: tmp,
+        tool_name: 'Task',
+        tool_input: {
+          subagent_type: 'mpl-phase-runner',
+          prompt: 'Execute phase-1.',
+        },
+        tool_response: '',  // empty final response after tool work
+        usage: { output_tokens: 0 },
+        metrics: { tools_used: 40, duration_ms: 28 * 60 * 1000 },
+      };
+      const r = JSON.parse(execFileSync('node', [HOOK_PATH], {
+        input: JSON.stringify(input),
+        encoding: 'utf-8',
+      }));
+      assert.equal(r.continue, true);
+      assert.match(r.systemMessage, /SUBAGENT RETURN ANOMALY/);
+      const state = readState(tmp);
+      // The anomaly is recorded …
+      assert.equal(state.subagent_return_anomalies.length, 1);
+      assert.equal(state.subagent_return_anomalies[0].agent_type, 'mpl-phase-runner');
+      // … but completed_todos must STAY at 0 — not auto-advance from the
+      // stale state-summary.md.
+      assert.equal(state.sprint_status.completed_todos, 0,
+        'phase-runner anomaly must NOT advance completed_todos');
+    } finally {
+      rmSync(tmp, { recursive: true, force: true });
+    }
+  });
 });
 
 /* ─── Stage A Phase 1.6c-i (PR #186 review fix): Bash gate routing ─── */
--- a/hooks/__tests__/mpl-gate-recorder.test.mjs
+++ b/hooks/__tests__/mpl-gate-recorder.test.mjs
@@ -393,6 +393,108 @@ describe('mpl-gate-recorder test-agent evidence', () => {
     assert.equal(ev.invalid_reason, 'missing_json_block');
   });
 
+  it('test-agent valid PASS JSON with anomaly is INVALID, not PASS (codex r4)', () => {
+    // Codex r4 on PR #218: a syntactically valid PASS JSON paired with a
+    // recorded subagent anomaly must NOT clear the verifier trust
+    // boundary. The anomaly is added to the issues list so invalid_reason
+    // stays non-null and isPassingTestAgentEvidence returns false.
+    const fenced = '```json\n' + JSON.stringify({
+      phase_id: 'phase-1',
+      verdict: 'PASS',
+      test_results: { total: 3, passed: 3, failed: 0, skipped: 0, pass_rate: 100 },
+      test_files_created: ['tests/phase-1.test.ts'],
+      commands_run: [{ command: 'npm test', exit_code: 0 }],
+      bugs_found: [],
+      a_item_coverage: [{ id: 'A-1', status: 'PASS' }],
+      s_item_coverage: [{ id: 'S-1', status: 'PASS' }],
+    }) + '\n```';
+    const ev = parseTestAgentEvidence({
+      phaseId: 'phase-1',
+      response: fenced,
+      anomaly: { type: 'zero_token_after_tools' },
+    });
+    assert.equal(ev.valid_json, true);
+    assert.equal(ev.verdict, 'INVALID');
+    assert.match(ev.invalid_reason, /subagent_anomaly:zero_token_after_tools/);
+    assert.equal(ev.subagent_anomaly_type, 'zero_token_after_tools');
+    assert.equal(isPassingTestAgentEvidence(ev), false);
+  });
+
+  it('phase-runner anomaly installs blocked_hook envelope (codex r4)', () => {
+    // Codex r4 on PR #218: phase-runner anomaly must structurally block
+    // progression — the orchestrator advances from PLAN.md state, not
+    // sprint_status, so a non-blocking systemMessage is insufficient.
+    const tmp = mkdtempSync(join(tmpdir(), 'mpl-gate-recorder-phase-runner-block-'));
+    try {
+      seedState(tmp);
+      const input = {
+        cwd: tmp,
+        tool_name: 'Task',
+        tool_input: {
+          subagent_type: 'mpl-phase-runner',
+          prompt: 'Execute phase-3.',
+        },
+        tool_response: '',
+        usage: { output_tokens: 0 },
+        metrics: { tools_used: 40, duration_ms: 28 * 60 * 1000 },
+      };
+      const r = JSON.parse(execFileSync('node', [HOOK_PATH], {
+        input: JSON.stringify(input),
+        encoding: 'utf-8',
+      }));
+      assert.equal(r.continue, true);
+      const state = readState(tmp);
+      assert.equal(state.session_status, 'blocked_hook',
+        'phase-runner anomaly must install a blocked_hook envelope');
+      assert.equal(state.blocked_by_hook, 'mpl-gate-recorder');
+      assert.match(state.block_code, /^phase_runner_/);
+      assert.equal(typeof state.retry_context, 'object');
+      assert.equal(state.retry_context.agent_type, 'mpl-phase-runner');
+    } finally {
+      rmSync(tmp, { recursive: true, force: true });
+    }
+  });
+
+  it('phase-runner anomaly does NOT clobber a pre-existing blocked_hook (codex r4)', () => {
+    const tmp = mkdtempSync(join(tmpdir(), 'mpl-gate-recorder-phase-runner-noclobber-'));
+    try {
+      seedState(tmp, {
+        session_status: 'blocked_hook',
+        blocked_by_hook: 'mpl-require-test-agent',
+        blocked_phase: 'phase-1',
+        blocked_artifact: 'state.test_agent_dispatched.phase-1',
+        block_code: 'missing_or_invalid_test_agent_evidence',
+        block_reason: 'pre-existing more specific block',
+        resume_instruction: 'follow existing recovery',
+        blocked_at: '2026-05-27T00:00:00.000Z',
+        retry_context: { from: 'earlier' },
+      });
+      const input = {
+        cwd: tmp,
+        tool_name: 'Task',
+        tool_input: {
+          subagent_type: 'mpl-phase-runner',
+          prompt: 'Execute phase-3.',
+        },
+        tool_response: '',
+        usage: { output_tokens: 0 },
+        metrics: { tools_used: 40, duration_ms: 28 * 60 * 1000 },
+      };
+      execFileSync('node', [HOOK_PATH], {
+        input: JSON.stringify(input),
+        encoding: 'utf-8',
+      });
+      const state = readState(tmp);
+      // The pre-existing more-specific block must survive.
+      assert.equal(state.blocked_by_hook, 'mpl-require-test-agent');
+      assert.equal(state.block_code, 'missing_or_invalid_test_agent_evidence');
+      // Anomaly is still recorded for visibility.
+      assert.equal(state.subagent_return_anomalies.length, 1);
+    } finally {
+      rmSync(tmp, { recursive: true, force: true });
+    }
+  });
+
   it('phase-runner anomaly does not advance sprint_status.completed_todos (codex r3)', () => {
     // Codex r3 on PR #218: when the phase-runner returns anomalous output
     // (empty / zero tokens after substantial tool work), the hook must not
--- a/hooks/__tests__/mpl-gate-recorder.test.mjs
+++ b/hooks/__tests__/mpl-gate-recorder.test.mjs
@@ -10,6 +10,7 @@ import {
   isPassingTestAgentEvidence,
   parseTestAgentEvidence,
   TEST_AGENT_EVIDENCE_PREVIEW_LIMIT,
+  TEST_AGENT_RESPONSE_PREVIEW_LIMIT,
 } from '../lib/mpl-test-agent-evidence.mjs';
 
 const __filename = fileURLToPath(import.meta.url);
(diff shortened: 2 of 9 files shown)


KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the evidence hardening is internally inconsistent and can both reject valid Task envelopes and accept outputs the new contract says are invalid.

Findings:
- [high] Content-array Task responses cannot clear test-agent recovery (hooks/lib/mpl-test-agent-evidence.mjs:29-36)
  `parseResponseJson` only recurses into wrapped responses when `content`, `output`, `response`, or `text` is a string. The new anomaly code explicitly treats `content: [{ type: 'text', text: ... }]` as a real Task final response, but this parser rejects the same shape as `missing_test_agent_fields`. A fenced PASS JSON inside that wrapper records INVALID evidence, so `mpl-require-test-agent`/`mpl-recover` can loop forever in runtimes that deliver Task output as content arrays.
  Recommendation: Normalize Task responses through the same final-text extractor before parsing, or add content-array handling to `parseResponseJson`; add an integration test where `{content:[{type:'text',text:'```json ... ```'}]}` records PASS and clears the matching block.
- [medium] Strict fenced JSON rule is not enforced (hooks/lib/mpl-test-agent-evidence.mjs:19-26)
  The agent prompt now says the final message must start/end with a ```json fence and prose outside it is invalid, but `firstJsonCandidate` accepts the first fenced block anywhere and then falls back to slicing from the first `{` to the last `}`. A response with prose before/after a PASS JSON object still becomes valid PASS evidence and can clear `blocked_hook`, so the recovery hardening is prompt-only for this failure mode.
  Recommendation: For string responses, require the entire trimmed response to match a single fenced JSON block; reject bare/prose-wrapped JSON with a distinct invalid reason and add regression tests for prose-before, prose-after, and bare JSON outputs.

Next steps:
- Unify response normalization between anomaly detection and test-agent evidence parsing.
- Make the parser enforce the same final-output contract documented in `agents/mpl-test-agent.md`.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at d27e29f.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the strict-fence gate still has a bypass that can accept non-fenced test-agent evidence as PASS.

Findings:
- [high] Object responses bypass strict fenced JSON enforcement (hooks/lib/mpl-test-agent-evidence.mjs:77-82)
  `parseResponseJson` returns valid for any object carrying `test_results` or `phase_id` before extracting assistant text and applying `strictFencedJsonCandidate`. That contradicts the new final-output contract: an object-shaped, non-fenced test-agent response with PASS fields is recorded as valid evidence and can satisfy `isPassingTestAgentEvidence`. I verified this path with a direct parser call returning `valid_json=true`, `verdict=PASS`, and passing evidence for a bare object payload.
  Recommendation: Route all Task responses through text extraction plus `strictFencedJsonCandidate`; only accept structured objects if they are wrapped in an explicit trusted internal envelope that cannot be produced by the test-agent final response.

Next steps:
- Add a regression test where `response` is a bare object with otherwise passing fields and assert it is INVALID under the strict-fence rule.
- Remove or narrow the direct-object acceptance path, then rerun the hook tests for gate recorder and require-test-agent.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 813eb5c.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new anomaly path can observe a failed phase-runner return while still advancing completion state, and the detector can also manufacture false anomalies from valid content-array responses.

Findings:
- [high] Phase-runner anomalies do not stop completion state from advancing (hooks/mpl-gate-recorder.mjs:360-408)
  The hook detects and records a Task/Agent return anomaly, but then the mpl-phase-runner branch still counts on-disk state-summary.md files, writes sprint_status.completed_todos, and returns continue. If a phase-runner crashes or returns an empty response after tool work while a stale or partial state-summary.md exists, MPL can mark the phase as completed and proceed; mpl-require-test-agent only blocks phases marked test_agent_required, so non-required phases can silently advance after a known-bad subagent return.
  Recommendation: Fail closed for mpl-phase-runner anomalies: do not update completed_todos on anomalous returns, and record a blocked_hook/paused recovery envelope or otherwise require explicit verification before allowing the phase transition to proceed.
- [medium] Empty alias fields can hide valid content and create false anomalies (hooks/lib/mpl-subagent-anomaly.mjs:40-53)
  extractFinalResponseText returns the first string-valued direct field even when it is empty, before checking response.content. A response shaped like { text: '', content: [{ type: 'text', text: '...' }] } is therefore treated as blank even though the final content exists. That records subagent_return_anomalies and emits re-dispatch recommendations for successful responses; for mpl-test-agent the evidence parser may still accept PASS, leaving contradictory PASS evidence with subagent_anomaly_type attached.
  Recommendation: Only accept non-blank direct text fields, or prefer the same content-array extraction logic used by parseTestAgentEvidence so anomaly detection and evidence parsing agree on what the final assistant text is.

Next steps:
- Add regression tests for anomalous mpl-phase-runner completions with stale state-summary.md and for response objects that contain empty direct aliases plus non-empty content arrays.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 7645497.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the anomaly handling is advisory in the phase-runner path and can still let anomalous verification evidence clear a test-agent block.

Findings:
- [high] Phase-runner anomalies do not structurally block progression (hooks/mpl-gate-recorder.mjs:403-414)
  On a phase-runner anomaly this code only skips updating `sprint_status.completed_todos`, appends telemetry, then returns `continue:true` with a `systemMessage`. The phase2 router advances from `PLAN.md` completion, not from `sprint_status`, so an empty/zero-token phase-runner return can still leave the orchestrator free to proceed if TODOs or stale artifacts already indicate completion. This is an inference from the unchanged phase-controller routing, but the changed code here does not create any blocked state or blocking hook decision.
  Recommendation: Make phase-runner anomalies create a real block, e.g. return a block decision or set `session_status:'blocked_hook'` with recovery context, and teach the phase router/recover flow to require explicit re-dispatch or operator override before transitioning.
- [high] Valid JSON test-agent output can pass despite a recorded anomaly (hooks/lib/mpl-test-agent-evidence.mjs:181-198)
  `parseTestAgentEvidence` sets an anomaly invalid reason in the base record, but when the JSON parses it builds `evidence` with `invalid_reason:null` and later can set `verdict:'PASS'`. Since `isPassingTestAgentEvidence` ignores `subagent_anomaly_type`, a `zero_token_after_tools` or similar anomaly can be recorded while the same evidence still clears the `mpl-require-test-agent` block. That contradicts fail-closed anomaly handling for the verifier trust boundary.
  Recommendation: Treat any supplied anomaly as non-PASS evidence before applying the normal PASS checks, or include `subagent_anomaly_type`/`anomaly.type` in the issue list so `invalid_reason` remains non-null and blocked-hook cleanup cannot run.

Next steps:
- Add regression tests for a phase-runner anomaly with completed PLAN/TODO state and for a syntactically valid test-agent JSON response accompanied by `zero_token_after_tools`.
- Wire anomaly recovery into the existing blocked-hook/recover path instead of relying on `systemMessage` only.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 30abb2e.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No ship: the new phase-runner anomaly block can be overwritten by the existing test-agent gate, so the structural stop this change relies on is not durable.

Findings:
- [high] Phase-runner anomaly block is lost when the test-agent gate also fires (hooks/mpl-require-test-agent.mjs:421-445)
  When an anomalous mpl-phase-runner return occurs, mpl-gate-recorder records a blocked_hook first, because hooks/hooks.json runs mpl-gate-recorder before mpl-require-test-agent. This code then reads state and unconditionally records its own missing_or_invalid_test_agent_evidence block for required phases. That overwrites the active phase_runner_* anomaly block. Once a later mpl-test-agent PASS lands, mpl-gate-recorder clears the mpl-require-test-agent block, leaving only a passive state.subagent_return_anomalies entry and allowing progress without the phase-runner anomaly ever being recovered or explicitly cleared. Impact: the intended fail-closed protection for empty/zero-token phase-runner completions is bypassed for required-test-agent phases, which are likely the higher-risk phases.
  Recommendation: Before calling recordBlockedHook, preserve any existing blocked_hook owned by another hook, especially mpl-gate-recorder phase_runner_* blocks. Either return a block decision with the existing block reason, or add a multi-block/pending-block mechanism so the test-agent block cannot erase the anomaly block. Add an end-to-end hook-order test covering: phase-runner anomaly -> require-test-agent block attempt -> test-agent PASS must not clear the phase-runner anomaly block.

Next steps:
- Patch mpl-require-test-agent to avoid overwriting non-owned blocked_hook envelopes.
- Add a regression test that invokes the hooks in hooks.json order for an anomalous required phase and proves the phase_runner_* block remains active after test-agent PASS.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 3fba4df.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: a transient phase-runner anomaly can leave MPL permanently paused with no implemented recovery path.

Findings:
- [high] phase_runner_* blocks have no clear or recovery path (hooks/mpl-gate-recorder.mjs:405-430)
  This branch creates a new blocked_hook owned by mpl-gate-recorder with block_code=phase_runner_<anomaly>, but the non-anomalous phase-runner path only updates sprint_status and never clears a prior mpl-gate-recorder block. The standard recovery tool also has no phase_runner_* handler, and mpl-require-test-agent explicitly avoids clearing this non-owned block. Result: after one empty/zero-token phase-runner return, even a later verified clean rerun or passing test-agent evidence can leave session_status=blocked_hook indefinitely, forcing manual state surgery.
  Recommendation: Add an explicit recovery path for phase_runner_* blocks: either a mpl-recover handler that validates the phase artifacts and clears the envelope, or gate-recorder logic that clears its own block on a later non-anomalous completion for the same phase. Add tests for anomaly -> clean rerun/recover -> block cleared.

Next steps:
- Implement and test a deterministic clear path for mpl-gate-recorder phase-runner anomaly blocks.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at a3f9db5.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new anomaly recovery path can clear a phase-runner anomaly block with evidence from the wrong phase, hiding an unrecovered failure and allowing state to advance.

Findings:
- [high] Clean phase-runner completion clears anomaly blocks without matching the blocked phase (hooks/mpl-gate-recorder.mjs:431-443)
  The self-clear path only checks that the active block is owned by `mpl-gate-recorder` and that `block_code` starts with `phase_runner_`. It does not verify that the non-anomalous completion belongs to `state.blocked_phase` or `retry_context.phase_id`, even though `phaseIdFromPrompt` was extracted earlier. A clean completion from a different phase, or an out-of-order parallel/manual phase-runner result, can clear the anomaly block for the failed phase and then update `sprint_status.completed_todos` from disk truth. The likely impact is a hidden unrecovered phase-runner failure and pipeline progression based on stale or unrelated artifacts.
  Recommendation: Only clear `phase_runner_*` blocks when the clean phase-runner completion matches the blocked phase, for example by requiring `phaseIdFromPrompt` to equal `state.blocked_phase` or `state.retry_context.phase_id`; if the blocked phase is unknown, keep the block and require explicit recovery.

Next steps:
- Add a regression test where a `phase_runner_*` block for `phase-1` survives a clean `mpl-phase-runner` completion for `phase-2`.
- Gate the self-clear condition on the blocked phase/retry context before updating `sprint_status.completed_todos`.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at a5ebfc0.

— from claude

KyubumShin:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the anomaly gate can falsely block documented background Task workflows before a subagent has actually returned final output.

Findings:
- [high] Background Task handles are treated as empty subagent failures (hooks/mpl-gate-recorder.mjs:363-423)
  `mpl-gate-recorder` runs anomaly detection on every Task/Agent response and installs a `blocked_hook` for `mpl-phase-runner` anomalies. Documented MPL flows use `run_in_background=true`, where the immediate Task result is a handle rather than final assistant text. Because the detector sees handle/metadata-only responses as blank and classifies any blank response as `empty_response`, a valid background phase-runner dispatch can be recorded as a failed return and freeze the pipeline before the actual completion is joined. This also risks poisoning test-agent evidence on background verifier dispatches.
  Recommendation: Before calling `detectSubagentReturnAnomaly` or parsing test-agent evidence, detect `tool_input.run_in_background`/handle-only responses and defer recording until the real completion event. Add regression tests for background `mpl-phase-runner` and `mpl-test-agent` Task payloads.

Next steps:
- Patch the recorder to distinguish background dispatch handles from completed subagent results.
- Add tests that a background Task handle does not create `subagent_return_anomalies`, `test_agent_dispatched`, or `blocked_hook` state.


— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at eb15c1c.

— from claude

KyubumShin:
### Issues at 4f2a84c

- **[logic] Background phase-runner handles still trigger the test-agent gate** `hooks/mpl-require-test-agent.mjs:387` — The diff teaches `mpl-gate-recorder` that `run_in_background` Task responses are only handle stubs and must not be treated as completion, but this hook still treats every `mpl-phase-runner` Task as a completed phase. For a required phase dispatched with `run_in_background: true`, the first handle-return PostToolUse event reaches this path, finds no PASS evidence in `state.test_agent_dispatched`, and records/returns a missing-test-agent block before the phase-runner has produced final output. That stalls background execution on a false completion condition. Suggested fix: Add the same `run_in_background` / `runInBackground` early allow used by `mpl-gate-recorder`, or otherwise gate this hook on a real phase-runner completion event rather than the initial background handle event.

— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 4f2a84c.

— from claude

KyubumShin:
### Issues at 205d866

- **[logic] Phase-runner anomaly block can race over pause/hang session states** `hooks/mpl-gate-recorder.mjs:421` — The new anomaly path installs a `blocked_hook` whenever `session_status !== 'blocked_hook'`. That includes `paused_budget` and `verification_hang`, which are active global stop states elsewhere in the system. A late or background phase-runner completion that returns anomalously after the run was paused/hung will overwrite that state with `blocked_hook`; `blocked_hook` is not a global Task-dispatch blocker, so the pause/hang guard can be bypassed. Suggested fix: Only install the anomaly block when `session_status` is null/active, or explicitly preserve stronger session states such as `paused_budget`, `paused_checkpoint`, `verification_hang`, and `cancelled`. Add a regression test seeding one of those states before an anomalous phase-runner completion.

— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 205d866.

— from claude

KyubumShin:
### Issues at 1a606ed

- **[data-integrity] Background Task completions lose verifier evidence** `hooks/mpl-gate-recorder.mjs:371` — The new background guard skips both anomaly detection and mpl-test-agent evidence parsing whenever the original Task input has run_in_background=true. The code assumes a later Task/Agent PostToolUse event will carry the real response, but the repo already documents background completions as UserPromptSubmit <task-notification> XML, and the executor docs allow background test-agent verification. In that path, the only PostToolUse event this recorder sees is the handle stub, so state.test_agent_dispatched[phase_id] is never populated even if the verifier later returns PASS. That loses the machine evidence Hard 2 and phase evidence latches depend on, causing required verification to be missing or unrecoverably blocked after successful background tests. Suggested fix: Do not skip solely on run_in_background. Skip only handle-stub responses, and add a recorder path for the actual background completion/join notification before relying on state.test_agent_dispatched. Cover this with an integration test that simulates the documented task-notification completion path.

— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 1a606ed.

— from claude

KyubumShin:
### Issues at 1f2cce1

- **[logic] String content completions are misclassified as background stubs** `hooks/mpl-require-test-agent.mjs:409` — The new handle-stub predicate exits before enforcement when run_in_background is true and the response has handle/taskId/task_id/id, but it only treats text/response/output strings and content arrays as usable payloads. A completion shaped like {id: "...", content: "phase complete"} therefore takes the stub path and returns continue before checking required mpl-test-agent PASS evidence. That lets a required phase-runner completion advance without state.test_agent_dispatched[phaseId] passing. This is an inferred response-shape risk, but it is supported by the parser in this diff accepting string content as final text. Suggested fix: Add string content to the usable-payload checks, preferably by sharing one helper with mpl-gate-recorder, and add a regression case for run_in_background responses that include id/handle plus content as a string.

— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 1f2cce1.

— from claude

KyubumShin:
✅ Merge OK — no blocking issues found at 4a94889.

— from codex

KyubumShin:
✅ Merge OK — no blocking issues found at 4a94889.

— from claude

Issue #225, opened 2026-05-28
[Exp22 follow-up #212] Brief generator agent + executor integration + cutover to block mode


KyubumShin:
Source: #212 MVP follow-up. PR #224 shipped the schema, validator, and PreToolUse gate in warn mode. The remaining producer + consumer integration lives here.

## Scope

1. **Brief generator** — either a new \`mpl-verification-brief-generator\` agent or an extension of \`mpl-seed-generator\` that writes \`.mpl/mpl/phases/{phase_id}/test-agent-brief.yaml\` per the schema in \`docs/schemas/test-agent-brief.md\`. Ownership decision is part of this issue.

2. **\`mpl-test-agent\` prompt update** — declare the brief as the primary execution contract. The agent should read \`.mpl/mpl/phases/{phase_id}/test-agent-brief.yaml\` first and use the brief's \`required_test_commands\` / \`a_item_coverage\` / \`s_item_coverage\` / \`forbidden_conditions\` / \`expected_evidence_shape\` to drive its run, instead of re-inferring from the decomposition.

3. **Executor dispatch** — \`commands/mpl-run-execute.md\` Step 4 should pre-load the brief path into the test-agent prompt instead of assembling scattered decomposition fields ad hoc.

4. **Cutover to block mode** — once the producer reliably writes a valid brief for every \`test_agent_required: true\` phase, ship \`.mpl/config/test-agent-brief-enforcement.json\` (or default the resolver) to \`"mode": "block"\` so the gate becomes load-bearing.

## Acceptance Criteria

- A required phase's brief is produced automatically (no manual artifact authoring needed for the common path).
- \`mpl-test-agent\` documents the brief as primary; existing scattered-field path becomes a fallback or is removed.
- Executor dispatch tests confirm the brief path is referenced in the dispatched prompt.
- Cutover commit flips the default enforcement mode to \`block\` and ships a migration note for existing workspaces.
- PR #224's gate semantics (warn vs block vs off via \`.mpl/config/test-agent-brief-enforcement.json\`) are not regressed; only the default flips.

## Non-Goal

Re-opening #212. The MVP closes the schema + validator + enforcement-gate gap; this issue closes the producer + consumer integration gap.

PR #224, opened 2026-05-28
Add test-agent brief schema + validation gate (#212 MVP)

KyubumShin:
## What

- **\`docs/schemas/test-agent-brief.md\`** (new): canonical schema for \`.mpl/mpl/phases/{phase_id}/test-agent-brief.yaml\` — the per-phase test-agent execution runbook. Required fields: \`phase_id\`, \`target_implementation_files\` (for code-bearing phases), A/S item coverage maps with concrete \`test_target\`, \`required_test_commands\`. Placeholder/echo/true commands and TODO/n/a targets are rejected.
- **\`hooks/lib/mpl-test-agent-brief.mjs\`** (new): \`validateBrief(text, { phaseId })\` with structured violation list. Minimal YAML peek (same style as \`mpl-gate-classify\`/\`mpl-phase0-artifacts\`).
- **\`hooks/mpl-require-test-agent-brief.mjs\`** (new): PreToolUse Task|Agent. Fires for \`subagent_type=mpl-test-agent\`. Reads decomposition.yaml for \`test_agent_required\`; reads brief; blocks dispatch on missing/invalid brief with copy-ready recovery hint.
- **\`hooks/hooks.json\`**: register the new hook.
- **\`docs/design.md\`**: hook table entry + count bump.

Tests in \`hooks/__tests__/mpl-require-test-agent-brief.test.mjs\` cover the validator + 5 integration scenarios:
1. missing brief for \`test_agent_required:true\` → block
2. invalid brief missing A/S coverage → block
3. invalid brief with placeholder command → block
4. valid brief → pass
5. \`test_agent_required:false\` phase → pass without a brief

## Why

Exp22 R12 / #212: decomposer responsibility concentration — it would have to ship full test-agent runbook context (commands, coverage, adversarial targets) on top of phase graph + contracts + type policy + E2E + probing hints. Splitting the brief out as its own artifact gives the role boundary a concrete schema-checked artifact instead of prose.

## Non-goal (follow-up)

This PR ships the schema + validator + enforcement gate. The brief PRODUCER — a new \`mpl-verification-brief-generator\` agent OR a \`mpl-seed-generator\` extension that writes the artifact — is intentionally deferred so producer and consumer can land independently. \`mpl-test-agent\` prompt updates that declare the brief as the primary contract, and executor dispatch changes that pre-load the brief, are also follow-ups.

## Design

Acceptance criteria from #212:
- ✅ A schema exists for \`test-agent-brief.yaml\`
[...]

PR #226, opened 2026-05-28
Auto-generate test-agent briefs + flip default to block (#225)

KyubumShin:
## Summary

- Closes the producer half of #225 (PR #224 shipped the schema + validator + gate in `warn` mode; this ships the producer + flips to `block`).
- New `writeTestAgentBriefs(cwd)` in `hooks/lib/mpl-decomposition-postprocess.mjs` — mechanical postprocess that derives `.mpl/mpl/phases/{phase_id}/test-agent-brief.yaml` from existing decomposition fields (no new agent dispatch, no new LLM call). Wired into the existing `mpl-decomposition-postprocess` hook so briefs regenerate on every `decomposition.yaml` write.
- Default enforcement mode flips `warn` → `block` in `hooks/mpl-require-test-agent-brief.mjs`. Operators can still opt back to `warn`/`off` via `.mpl/config/test-agent-brief-enforcement.json`.

## Approach

The decomposer already emits every field the test-agent brief needs:

| Brief field                  | Source                                |
|------------------------------|---------------------------------------|
| `target_implementation_files`| `impact.modify[]`                     |
| `interface_contracts`        | `interface_contract.produces[]`       |
| `a_item_coverage`            | `verification_plan.a_items[]`         |
| `s_item_coverage`            | `verification_plan.s_items[]`         |
| `probing_targets`            | `probing_hints[]`                     |
| `required_test_commands`     | `phase_lang` → npm/pytest/cargo/go    |

So the third path (mechanical postprocess) was chosen over the listed candidates (new agent / `mpl-seed-generator` extension). Keeps decomposer responsibility surface unchanged; no new dispatch in any pipeline.

One subtle fix: `decomposition.yaml` has `- id:` lines at the phase level AND inside `a_items:` / `s_items:` lists. The naive split that the existing parser uses for other postprocess derivations would have treated each A/S item as its own phase. `extractPhaseBlocks` is indent-aware — it locks in the phase-level indent from the first `phases:` array item and only splits on `- id:` occurrences at that exact indent.

## Still-deferred (separate follow-up)

- `mpl-test-agent` system prompt update declaring the brief as primary execution contract (consumer-side).
- `commands/mpl-run-execute.md` Step 4 to pre-load the brief into the dispatched prompt (consumer-side).

The producer + cutover is the load-bearing half. Consumer-side prompt changes can land independently now that briefs are guaranteed.

## Test plan

- [x] `node --test hooks/__tests__/mpl-decomposition-postprocess.test.mjs` — 16/16 pass, includes new tests for: brief auto-written per required phase, `test_agent_required: false` skips brief, idempotency, implicit-default, python → pytest derivation.
- [x] `node --test hooks/__tests__/mpl-require-test-agent-brief.test.mjs` — 19/19 pass, scenario 1 updated for new block-default, new scenario 1b covers explicit warn-mode override.
- [x] Full suite `node --test "hooks/__tests__/*.test.mjs"` — 1254/1254 pass.


KyubumShin:
Out-of-scope: The brief is a derived artifact by design — `docs/schemas/test-agent-brief.md` and the auto-generation header in every brief file explicitly state "Manual edits will be overwritten on the next postprocess run — change the source decomposition instead." The "data loss" codex flagged is the documented intentional contract of the new producer. Hand-authored briefs during the warn-default window (#224) were always a transitional workaround, not a feature to preserve.

The PR `## Why` does not assert "hand-authored briefs preserved" as an invariant; the opposite is the design. Per Convention §8, [data-integrity] applies when an invariant explicitly stated in PR `## Why` is now violated — that condition is not met here.

Findings [security] and [contract-break] addressed in 25f8916.

— from author

PR #219, opened 2026-05-27
Guard gate evidence command families on state writes (#209)

KyubumShin:
## What

- `hooks/lib/mpl-gate-classify.mjs` (new): shared Bash→gate-family classifier, extracted from `mpl-gate-recorder` so the state invariant uses the same pattern lists. Adds a head-command denylist (`git`, `gh`, `echo`, `cat`, `docker`, ...) so commands whose head is never gate evidence get classified as null even when arguments happen to contain a family keyword.
- `hooks/lib/mpl-state-invariant.mjs`: new `VIOLATION_IDS.GATE_COMMAND_FAMILY_MISMATCH` (\`I12\`). Fires on STATE_WRITE; walks `state.gate_results` + `state.release.gate_results` and rejects entries whose `command` doesn't belong to the slot's family. Violation message names the offending key + command verbatim.
- `hooks/mpl-gate-recorder.mjs`: refactored to import the shared `classifyGateCommand`. Same patterns, now also rejects head-denylist commands at the recorder level.
- Tests cover valid Hard 1/2/3 commands, `git commit` in H2/H3, release-scoped writes, legacy-booleans-true with invalid structured command, wrong-family commands (build in H2 slot), STOP-trigger skip, and the null-entry edge cases.

## Why

Exp22 R13 (V1 harness coverage gap analysis) found post-run state with structured-looking gate entries whose commands were the wrong evidence: `hard2_coverage.command` and `hard3_resilience.command` held `git commit ...`. Legacy `hard{1,2,3}_passed: true` booleans made the state appear complete despite the family mismatch.

This guard closes that gap on the write boundary so manual `.mpl/state.json` patches cannot pass unrelated commands as gate evidence; existing recorder-produced evidence remains valid because the classifier patterns are unchanged.

## Design

Implements all of the #209 acceptance criteria. Bucket [data-integrity] under \`/pr\` Convention §8.

## Agent Review Status

This PR is gated on **2 unique agent reviews** (footers \`— from <agent>\`).

- [ ] from claude
- [ ] from codex

Closes #209.

PR #231, opened 2026-05-28
Reject composite/pipe/subshell at strict gate classifier (#220)

### LINKED ISSUE #220: [Exp22 follow-up #209] Strict classifier — additional bypass shapes to harden
Source: codex review iterations on PR #219 (rounds 1–7). The strict classifier in \`hooks/lib/mpl-gate-classify.mjs\` has shipped with a head allowlist plus denylist + wrapper/flag fail-closed. Each round of review surfaced a narrower bypass shape; rounds 1–7 are all addressed in #219. This issue tracks additional shapes that codex might surface in future iterations on the same surface, plus general defense-in-depth refinements.

## Known bypass shapes already addressed in PR #219

- r1: path-qualified \`/usr/bin/git ...\` → basename reduction
- r3: shell wrapper (\`bash -lc\`) → denylist
- r4: eval / subshell (\`eval\`, \`(...)\`, backticks, \`$(...)\`) → denylist + subshell-opener strip
- r5: wrapper + flag (\`sudo -E git ...\`, \`env -u FOO ...\`) → fail-closed on flag after prefix
- r7: script-interpreter eval (\`node -e\`, \`python -c\`) → switch to strict allowlist

## Candidate further hardening

- Inline composite commands separated by \`;\` / \`&&\` / \`||\` — currently the head is the first command, but the rest of the regex may match a downstream command.
- Heredoc-piped commands (\`cat <<EOF | sh\`).
- Pipes (\`cat fake-log | grep e2e\`) — full-string regex still sees keywords.
- Allowlist expansion when MPL grows new test runners (e.g. \`bun test\`, \`deno test\`).

## Non-goal

Not re-opening #209. Its acceptance criteria are met by the merged PR. This issue is defense-in-depth for the strict classifier surface.

KyubumShin:
## Summary

- Closes #220. Strict `classifyGateCommand` now rejects composite shells (`;`, `&&`, `||`), pipes (`|`), and command substitution (backticks, `$(`) — closing the empirical bypass where `npm test; git commit -m e2e` classified as hard3.
- Loose `classifyRecordedCommand` (recorder path) is unchanged; legitimate wrapper invocations still record correctly.
- Two new tests cover composite-with-keyword-downstream and pipe / substitution cases.

## Why

Empirically reproduced: `npm test; git commit -m e2e` had its head extracted as `npm` (in the allowlist), passed `STRICT_GATE_HEAD_ALLOWLIST.has(head)`, then `matchFamilyRegex(command)` ran against the whole composite and matched `e2e` from the commit message — classifying as `hard3_resilience`. A manual `state.gate_results` write could mark a non-e2e command as e2e gate evidence.

Per #220's scope, this is the composite/pipe variant of the same masquerade family that PR #219 hardened against (path-qualified, shell wrappers, subshell/eval, wrapper-with-flag, script-interpreter eval).

## Approach

Same fail-closed pattern PR #226 (#225 brief command guard) adopted: regex-reject the dangerous shell metacharacters anywhere in the command before head extraction. No quote-state parsing — manual gate evidence is a single command and never needs `;`/`&&`/`||`/`|`/backticks/`$(`.

## Test plan

- [x] `node --test hooks/__tests__/mpl-state-invariant.test.mjs` — 83/83 pass, including 2 new #220 regression tests.
- [x] Full hook suite — 1281/1281 pass.

### DIFF
--- a/hooks/__tests__/mpl-state-invariant.test.mjs
+++ b/hooks/__tests__/mpl-state-invariant.test.mjs
@@ -351,6 +351,45 @@ describe('I12 gate command-family mismatch (Exp22 R13 / #209)', () => {
     }
   });
 
+  it('#220: composite shell forms (semicolon / && / ||) are rejected at the strict level', () => {
+    // Empirical bypass: `npm test; git commit -m e2e` has head `npm`
+    // (allowlisted) but the family regex sees `e2e` downstream and
+    // classifies as hard3. Manual gate evidence is a single command;
+    // composite/pipe/subshell shells fail closed.
+    for (const cmd of [
+      'npm test; git commit -m e2e',
+      'npm test && git commit -m playwright',
+      'npm run build || echo failed',
+      'npm test; touch e2e.json',
+      'tsc --noEmit; echo "playwright"',
+    ]) {
+      const r = checkInvariants({
+        gate_results: {
+          hard2_coverage: gateEntry(cmd),
+        },
+      }, { cwd: tmp, trigger: TRIGGERS.STATE_WRITE });
+      const v = r.violations.find((x) => x.id === VIOLATION_IDS.GATE_COMMAND_FAMILY_MISMATCH);
+      assert.ok(v, `composite shell form must be rejected at strict level: ${cmd}`);
+    }
+  });
+
+  it('#220: pipes and command substitution are rejected at the strict level', () => {
+    for (const cmd of [
+      'npm test | tee output.log',
+      'echo "$(npm test)"',
+      'cat fake-log | grep e2e',
+      'npm test `echo --quiet`',
+    ]) {
+      const r = checkInvariants({
+        gate_results: {
+          hard2_coverage: gateEntry(cmd),
+        },
+      }, { cwd: tmp, trigger: TRIGGERS.STATE_WRITE });
+      const v = r.violations.find((x) => x.id === VIOLATION_IDS.GATE_COMMAND_FAMILY_MISMATCH);
+      assert.ok(v, `pipe / command substitution must be rejected: ${cmd}`);
+    }
+  });
+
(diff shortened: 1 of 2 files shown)


KyubumShin:
Codex r6 findings (round 7) are both out-of-scope for #220:

Out-of-scope: Both r7 findings push beyond #220's "strict classifier additional bypass shapes" charter into deeper recorder semantics.

Finding 1 ([contract-break] round-trip property) — my r5 commit message overpromised. The strict / recorder asymmetry is INTENTIONAL: strict rejects wrapper heads (docker, bash -lc, kubectl exec) that the recorder accepts as legitimate test wrappers. I12 uses strict to BLOCK manual masquerade, not to re-validate recorder writes. The fix is documenting this in code comments, not making strict accept docker/bash. Made the doc update in 5034694; the actual fix for "I12 should not re-validate recorder evidence" requires either a write-source marker or a separate invariant trigger — design decision tracked as #232.

Finding 2 ([data-integrity] composite exit code) — this is about RECORDER semantics (exit code captured from composite shell vs. leading gate command), not classifier bypass. `npm test || true` has the recorder storing the overall exit_code 0 from `true` while classifying as hard2 from the leading `npm test`. Real correctness gap, but fixing it requires either rejecting control operators in the recorder (consistent with strict) or capturing PIPESTATUS / using `set -o pipefail` — both are design decisions for the recorder, not #220's classifier surface. Tracked as #232.

Per the standing instruction to split scope expansion into follow-up issues rather than absorbing into the current PR, both are deferred. PR #231 closes the documented bypass shapes in #220's charter (composite + redirect + comment + process substitution + braces + newline) and ships the unified trim canonicalization that lets the strict and recorder paths share masquerade defenses.

— from author

Issue #238, opened 2026-05-28
[3-layer audit] Soft-signal telemetry surface for prompt-only quality rules (HA-01 / probing hints / seed invention / etc.)


KyubumShin:
Source: post-Exp22 3-layer pattern audit (see PR #233 / `docs/findings/2026-05-28-prompt-gate-restore-divergence.md` §A2, A4, A5, A6, A7, A8, A9, A10).

## Problem

Several MUSTs / SHOULDs are documented only in agent prompts. They're real quality concerns but blocking on them would be too aggressive — pattern matches are heuristic, false-positive rate is meaningful, and the rule is about coordination quality rather than artifact correctness.

Right now there's no middle path: the rule either blocks (too rigid) or is invisible (current state). The intended fix is a **telemetry surface** — structured records in a learning log so the violations are countable across runs, surfaced in `mpl-doctor` / `mpl-trace`, and reviewable without blocking the pipeline.

## Rules in scope

Per `docs/findings/2026-05-28-enforcement-relaxation-plan.md` §A:

| # | Rule | Recommended tier |
|---|---|---|
| A2 | HA-01 vague delegation prompt ("이전 결과 참고", "알아서 판단") | warn (systemMessage) + telemetry |
| A4 | `mpl-validate-output` JSON-fence rule | warn (already) + structured telemetry |
| A5 | `test_command: "TODO(integration-ci)"` placeholder | warn during iteration, block on finalize |
| A6 | Probing hints → ≥1 adversarial test | warn (zero coverage block-eligible) |
| A7 | Seed Generator "No invention" (ambiguity_notes usage) | warn |
| A8 | HA-02 BEGIN/END region mirror divergence | telemetry only |
| A9 | Retry 2 "must not do" reflection | telemetry only |
| A10 | Interviewer comparison table | warn (already substring-checked) |

## Scope

Single shared mechanism, multiple rule integrations:

1. **Telemetry log path**: `.mpl/mpl/quality-signals.jsonl` — append-only, one JSON record per signal, schema `{ rule, severity, ts, phase, agent, evidence: {...} }`.
2. **mpl-doctor surface**: summarize signal counts per rule per run; flag rules whose signal rate exceeds a threshold.
3. **Per-rule integration**: each rule above gets a detection point at the right hook (where the artifact is observable). Detection emits the telemetry record, optionally also a warn-level systemMessage, never blocks (except A5's finalize-time elevation).
[...]

Issue #236, opened 2026-05-28
[3-layer audit] Tighten mpl-write-guard for protected paths (decomposition.yaml + mpl-cancel safe-cleanup hole)


KyubumShin:
Source: post-Exp22 3-layer pattern audit (see PR #233 / `docs/findings/2026-05-28-prompt-gate-restore-divergence.md` §A1 + §A3).

## Problem

Two prompt-only rules govern destructive write/delete paths that `hooks/mpl-write-guard.mjs` actually permits. The gates are documented in agents/skills but the mechanical guard is missing.

### A1 — Orchestrator must NOT directly Write/Edit `.mpl/mpl/decomposition.yaml`

`commands/mpl-run-decompose.md:16, 24, 188` declares:

> The orchestrator MUST NOT Write or Edit `.mpl/mpl/decomposition.yaml` under any circumstances.

But `hooks/mpl-write-guard.mjs:54-62` explicitly allows `.mpl/` paths. Content gates (`mpl-require-phase-contract-graph.mjs`, `mpl-artifact-schema.mjs`) only check the presence of the string `generated_by: mpl-decomposer` in the file content — they don't verify the actual writing agent. An orchestrator (or any non-decomposer agent) can hand-write `decomposition.yaml` with the marker string and bypass the rule.

Impact: corruption of the single source of phase truth. Cross-phase contracts, MVP scope, and gate command families all derive from this file.

### A3 — `mpl-cancel` must NEVER delete protected paths

`skills/mpl-cancel/SKILL.md:104, 108, 110, 113` lists protected directories that the skill is NOT allowed to remove:
- `.mpl/mpl/**` (decomposition artifacts)
- `.mpl/contracts/*.json` (phase contracts)
- `docs/learnings/` (cumulative project memory)
- `.mpl/memory/` (cross-session auto-memory)

But `hooks/mpl-write-guard.mjs:111` actively *allowlists* `rm -rf .mpl` as a "safe cleanup" Bash pattern. The dangerous-bash check therefore PERMITS the exact destructive operation the skill forbids. The "Regression guard" the SKILL promises at line 113 has no mechanical counterpart.

Impact: a drifted mpl-cancel (or any agent invoking `rm -rf .mpl/...`) can wipe project memory + decomposition history with no gate firing.

## Scope

Both are write-guard tightening. Single hook, two related changes:

[...]

PR #249, opened 2026-05-29
Tighten mpl-write-guard for protected paths (#236 A1 + A3)

KyubumShin:
## What

Two write-guard tightenings per #236 acceptance criteria:

- **A1** — `mpl-write-guard.mjs` now requires `mpl-decomposer` subagent identity for `.mpl/mpl/decomposition.yaml` writes. The hook records a state dispatch flag when the orchestrator calls `Agent(subagent_type='mpl-decomposer')`; subsequent Write/Edit on decomposition.yaml is allowed only while that flag is active and within a 30-min TTL. Orchestrator hand-writes (or any non-decomposer agent) → block + envelope (`decomposition_writer_violation`).
- **A3** — `.mpl` removed from the safe-cleanup allowlist. New `matchesProtectedDelete` hard-blocks `rm`-family commands targeting `.mpl/mpl`, `.mpl/contracts`, `.mpl/memory`, `docs/learnings` (per `skills/mpl-cancel/SKILL.md`). Operator override via `MPL_FORCE_PURGE=1` env var.

`docs/design.md` Hook System table updated to reflect the new Task/Agent matcher.

## Why

Per `docs/findings/2026-05-28-prompt-gate-restore-divergence.md` §A1 + §A3 — both prompt-only rules previously had no mechanical guard. The decomposition.yaml content check only verified a `generated_by:` marker string (trivially forgeable). The dangerous-bash check actively allowlisted `rm -rf .mpl` as "safe cleanup", directly contradicting the mpl-cancel SKILL.

## Acceptance criteria

- [x] Orchestrator Write/Edit of decomposition.yaml is blocked at PreToolUse with an envelope referencing mpl-decomposer as the only allowed writer.
- [x] `rm -rf .mpl/mpl/phases/phase-1/` is blocked at PreToolUse with an envelope explaining the protected-path rule.
- [x] Legitimate `mpl-decomposer` subagent dispatch → `Write decomposition.yaml` still works (dispatch flag active path).
- [x] `MPL_FORCE_PURGE=1` override works for legitimate cleanup.
- [x] Tests cover both block cases + the legitimate-cancel pass case + the decomposer-write pass case.
- [x] Existing tests stay green (1399 → 1411).

## Out of scope

- **A2** (HA-01 vague delegation prompt heuristic) — Non-Goal per #236 spec; needs design pass on false-positive tolerance.
- Other A-category warn/telemetry items — separate issue (#238).

## Test plan

- [x] `node --test hooks/__tests__/*.test.mjs` — 1411/1411 pass.
- [x] `node --test hooks/__tests__/mpl-issue-236-write-guard-tighten.test.mjs` — 12/12.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -456,7 +456,7 @@ All Discoveries are recorded in `.mpl/discoveries.md`.
 |----|--------|------|------------|
 | `mpl-compaction-tracker` | PreCompact | Track compaction events and create checkpoints (F-31). | v0.13.x baseline |
 | `mpl-auto-permit` | PreToolUse | Apply learned safe permission decisions (F-34). | v0.13.x baseline |
-| `mpl-write-guard` | PreToolUse: Edit/Write/MultiEdit/Bash | Warn or block unsafe direct source edits and dangerous shell commands while MPL is active. | v0.13.x baseline; MultiEdit/MCP hardening v0.18.1 |
+| `mpl-write-guard` | PreToolUse: Edit/Write/MultiEdit/Bash/Task/Agent | Warn or block unsafe direct source edits and dangerous shell commands; #236: protect mpl-cancel SKILL paths from `rm -rf` and require mpl-decomposer subagent identity for decomposition.yaml writes. | v0.13.x baseline; MultiEdit/MCP hardening v0.18.1; protected-path + decomposition-writer v0.18.5 |
 | `mpl-bash-timeout` | PreToolUse: Bash | Enforce timeout budgets on build, lint, test, and verification commands. | v0.18.1 |
 | `mpl-state-invariant` | PreToolUse: Task/Agent/Edit/Write/MultiEdit; Stop | Validate state schema, gate evidence, pause/block status, and completion invariants before state can drift. | v0.18.1; I10/I11 recovery v0.18.4 |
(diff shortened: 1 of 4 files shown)


KyubumShin:
✅ Merge OK — no blocking issues found at 4f11eae.

r28 static review covered the two latest commits:
- **31a1850** — hoisted `normalizeShellCommand` ABOVE the destructive-verb gate so quote/backslash-concat VERB tokens (`r"m"`, `g"zip"`, `g\zip`) now trip the gate. The path-matching code below operates on the same normalized string.
- **4f11eae** — gzip keep-flag tokenization. Only standalone `-k` / `--keep` / `--keep=…` / combined short-bundle `-Xk…` tokens (after the compressor verb and before any literal `--`) count as the keep flag. Filename substrings containing `-k` no longer suppress destructiveness.

Adversarial probes tried (all either correctly blocked or non-exploitable in a real shell):
- `gzip foo bar -k` — `-k` is a global flag applying to all operands, suppression correct
- `gzip -- -k` — `--` terminates options, `-k` is operand → destructive
- multi-segment `gzip foo;gzip -k bar` — `.some()` on segments catches the unkeyed first segment
- `gzip $'\x2dk\x20.mpl/mpl/foo'` — bash parses as one quoted arg, gzip ENOENTs → not exploitable
- `gzip -k=NO` — over-blocks (correct, real gzip rejects), no destructive write

— from claude

Issue #245, opened 2026-05-29
Follow-up to #240 / PR #244: residual same-class I12 forgery surface (subcommand argv keywords + subshell expansion)


KyubumShin:
PR #244 went 10 rounds of cross-model adversarial review (codex + claude). Both reviewers ultimately posted Merge OK at d7ed9b7. The same-class forgery surface defended through r3–r9 (eval flags, quoting forms, runner-script positions, global flags, cargo/go subcommand allow-list) is now structurally closed.

Two residual same-class instances surfaced as **advisory** in r10 (concrete repro, but acknowledged by the reviewer as the same class repeatedly defended, surviving only on the subcommands the r9 allow-list intentionally lets fall through to `matchFamilyRegex`).

Tracking as a follow-up so PR #244 (which is at its core the #240 config-knob A1–A5 PR) can land. The classifier hardening was already scope expansion driven by adversarial review.

## (1) Trailing keyword arg promotes the entry to hard3

```js
classifyGateCommand('npm test playwright')   // → 'hard3_resilience' (should be hard2)
classifyGateCommand('cargo test playwright') // → 'hard3_resilience' (should be hard2)
classifyGateCommand('go build e2e')          // → 'hard3_resilience' (should be hard1)
classifyGateCommand('pnpm test playwright')  // → 'hard3_resilience'
```

`matchFamilyRegex` checks `HARD3_PATTERNS` first; an attacker-controlled trailing arg containing the e2e keyword wins over the anchored npm/cargo/go family pattern.

Suggested fix: when classifying via the npm-family / cargo / go fall-through, classify only by the anchored family regex for that head's subcommand — not by HARD3 keyword priority across the whole canonical. Effectively: pick the family from the subcommand, not from whichever keyword appears anywhere.

## (2) Subshell expansion inside the canonical

```js
classifyGateCommand('npm test \$(echo playwright)')  // → 'hard3_resilience'
classifyGateCommand('npm test \`echo playwright\`')  // → 'hard3_resilience'
```

`stripNonExecutedSuffix` (`hooks/lib/mpl-gate-classify.mjs:~677`) cuts at `;`, `|`, `&&`, etc. but NOT at `\` (backtick), `\$(`, `<(`, `>(`. PR #220/#231 r1 hard-rejected `(` and backticks; r5's unified strict + recorder trim approach dropped that rejection.

[...]

Issue #232, opened 2026-05-28
[Exp22 follow-up #220] Recorder semantic gaps — exit code vs leading command + strict/recorder allowlist divergence


KyubumShin:
Source: codex r7 review on PR #231 — two findings flagged as scope expansion beyond #220's "strict classifier additional bypass shapes" charter.

### 1. Recorder records composite exit code while classifying only the leading command

After PR #231 (#220), the recorder cuts at first `;`/`&&`/`||`/`|`/`&`/newline/redirect/comment and classifies only the leading simple command. But the recorder still stores the OVERALL Bash shell exit code from the composite. So:

- `npm test || true` → classifies as hard2 (leading `npm test`), exit_code records as `0` from `true` even when `npm test` failed.
- `npm test; true` → same shape.
- `npx playwright test | tee output.log` → classifies as hard3, exit_code is `tee`'s exit (typically 0) — not `npx playwright test`'s exit.

False PASS gate evidence is possible. Codex's recommendation: reject control operators in the recorder path UNLESS the hook can capture the leading gate command's actual status, or require `set -o pipefail` / `PIPESTATUS` for pipelines.

### 2. Strict / recorder allowlist divergence breaks the round-trip property

PR #231 r5 attempted to unify the two paths via `stripNonExecutedSuffix`. But the recorder accepts wrapper heads like `docker`, `bash`, `kubectl exec` while strict's allowlist only contains test runners (`npm`, `pnpm`, `yarn`, `npx`, `cargo`, `go`, `tsc`, `eslint`, ...). So a recorder write of `docker compose run app npm test` classifies as hard2 at write time, but a later I12 STATE_WRITE check re-classifies the same stored string as `null` (head `docker` not in strict allowlist) — manual-write rules wrongly apply to recorder evidence.

This is partly by design (strict purposely tighter for manual writes), but the I12 invariant currently treats both writes identically. Two possible directions:

- Document the asymmetry and have I12 skip re-validation of recorder writes (need a marker on the entry, or use a different invariant trigger).
- Make recorder canonicalize the wrapper away (e.g. record `docker compose run app npm test` as `npm test`) so strict round-trips. Loses provenance.

## Scope

- Decide the source-of-truth for stored gate evidence: recorder-classified-at-write-time vs strict-revalidated-at-read-time.
- For the exit_code question: implement PIPESTATUS / `set -o pipefail` capture, OR reject control operators / pipelines in the recorder path (consistent with strict).

## Acceptance Criteria

[...]

Issue #230, opened 2026-05-28
[Exp22 follow-up #214] Canonical failure_code separation + per-event failure cause requirement


KyubumShin:
Source: codex r6 review on PR #229 — two findings flagged as scope expansion beyond #214's original "validate explanation references at least one computed reason" charter.

### 1. Free-form failure_reason values can paraphrase-bypass the canonical-vocabulary gate

`event.failure_reason` is currently a free-form `err.message` string from the executor. The aggregator unions it into `rejection_reasons`. A failure like `failure_reason: "worker dispatch failed"` becomes a "canonical" token that the explanation can repeat verbatim — even though the words `"worker dispatch failed"` are exactly the kind of paraphrase the canonical-vocabulary gate (#214) is meant to block when an operator types them by hand.

### 2. Union reason can mask a parallel_failed runtime cause

`parallel_failed` events can carry BOTH `rejection_reasons_by_phase` (pre-attempt deferred-phase block reasons) AND `failure_reason` (runtime cause). The new content-strength check accepts any token from the union. A failed wave with by-phase `file_overlap` + runtime `worker_dispatch_error` can satisfy the gate by naming only `file_overlap`, hiding the actual runtime failure.

## Scope

Both findings push toward the same design:
- Add a canonical `failure_code` enum field to `parallel_failed` events (separate from the free-form `failure_reason` message string).
- Aggregator surfaces `failure_codes` distinct from `rejection_reasons`.
- Finalize gate requires each `parallel_failed` wave's canonical `failure_code` to appear in `no_parallel_explanation`, separately from the union-reason check.
- Producer docs (commands/mpl-run-execute.md / mpl-run-finalize.md) define the canonical `failure_code` allowlist.

Alternative shape: validate every `rejection_reasons` element against a strict canonical-token grammar (snake_case, ≤ 3 words, length ≤ 32, etc.) and reject free-form prose at the executor boundary — keeps the gate's vocabulary clean without introducing a new field.

## Acceptance Criteria

- A `parallel_failed` event with `failure_reason: "worker dispatch failed"` either:
  - (a) requires a separate canonical `failure_code` token in `no_parallel_explanation`, OR
  - (b) is rejected at the executor boundary as a non-canonical reason.
- A failed wave with both by-phase `file_overlap` AND runtime cause requires BOTH tokens in the explanation.
- Existing PR #229 tests stay green.

## Non-Goal

[...]

PR #261, opened 2026-05-30
Close non-runner head trailing-keyword forgery in I12 classifier (#256)

### LINKED ISSUE #256: Follow-up to #245: extend synthesized-canonical pattern to non-runner gate heads (tsc/eslint/vitest/jest/mocha/...)
Pre-existing on origin/main before PR #255 (#245). Surfaced as advisory by claude r1 on PR #255. Not a regression from #245, but the same structural class — recording here so the synthesized-canonical pattern can be generalized in a focused PR.

### (A) Trailing keyword arg on non-runner gate heads

\`hooks/lib/mpl-gate-classify.mjs\` non-runner branch falls straight to \`matchFamilyRegex(safeCanonical)\` which scans HARD3 keywords across the whole canonical. Trailing positional args containing an e2e keyword promote the family.

Repros:
\`\`\`js
classifyGateCommand('tsc playwright')      // → 'hard3_resilience'  (should be 'hard1_baseline')
classifyGateCommand('eslint cypress')      // → 'hard3_resilience'  (should be 'hard1_baseline')
classifyGateCommand('vitest playwright')   // → 'hard3_resilience'  (should be 'hard2_coverage')
classifyGateCommand('jest cypress')        // → 'hard3_resilience'  (should be 'hard2_coverage')
classifyGateCommand('mocha e2e')           // → 'hard3_resilience'  (should be 'hard2_coverage')
\`\`\`

Proposed fix: For non-runner heads in \`STRICT_GATE_HEAD_ALLOWLIST\` whose family is determined by the binary name itself (tsc/eslint/vitest/jest/mocha/ruff/mypy/...), synthesize a canonical of just the head before running matchFamilyRegex. Positional test selectors / file paths no longer leak into the keyword scan.

### (B) ANSI-C quoting / parameter expansion on non-runner heads

Same head set; \`\$'...'\` (ANSI-C) and \`\${...}\` (parameter expansion) wrap arbitrary text that the family regex matches as if it were a positional arg.

Repros:
\`\`\`js
classifyGateCommand("tsc \$'playwright'")       // → 'hard3_resilience'
classifyGateCommand('tsc \${FOO}playwright')    // → 'hard3_resilience'
classifyGateCommand('eslint \$"cypress"')       // similar shape
\`\`\`

Proposed fix: extend the (A) synthesized-canonical strategy — once the head is known, drop everything else before family classification on non-runner heads.

## Why follow-up

Per #245's bounded scope (the two PR #244 r10 paths). Both items are pre-existing on main and structurally same-class as #245 (1); generalizing the synthesized-canonical pattern to non-runner heads is a focused follow-up rather than scope expansion on #245.

## Test plan

- regression tests for (A) and (B) using the repros above; expected classifications listed for each head.
- existing #245 / #244 / #240 tests must continue to pass.

KyubumShin:
## Summary

Closes #256.

Extends the PR #244 / #245 synthesized-canonical pattern from runner heads (npm/pnpm/yarn/cargo/go/npx/pnpx) to the remaining `STRICT_GATE_HEAD_ALLOWLIST` heads whose binary name IS the family signal — `tsc`/`eslint`/`vitest`/`jest`/`mocha`/`ruff`/`mypy`/`playwright`/`cypress`/`wdio`.

## What

Before this PR, the non-runner branch of `classifyGateCommand` (after `classifyRunnerHead` returns `undefined`) fell straight to `matchFamilyRegex(safeCanonical)`, which scanned HARD3 keywords across the whole canonical. Two structurally-related leak shapes:

- **(A) Trailing positional keyword arg:** `tsc playwright`, `eslint cypress`, `vitest playwright`, `jest cypress`, `mocha e2e` all promoted to `hard3_resilience` via the trailing keyword, even though the head's binary name already determined the family.
- **(B) ANSI-C / parameter expansion / locale-translation quoting:** `tsc $'playwright'`, `tsc ${FOO}playwright`, `eslint $"cypress"` shared the same forgery surface because the wrapper text reached the family regex unmodified.

Fix: after `classifyRunnerHead` returns `undefined`, if the head is a `STRICT_HEAD_AS_FAMILY_SIGNAL` member, run `matchFamilyRegex(head)` — only the head itself, no trailing args. Both (A) and (B) close in the same change because the post-head text never reaches the regex.

## Why

Same I12-evidence-forgery class as PRs #219 / #244 / #245. Manual `state.gate_results` writes must classify by their actual binary, not by an attacker-controllable trailing-arg or quoting wrapper. Each head in the new set already appears as a `\b<head>\b` literal in exactly one of `HARD1/HARD2/HARD3_PATTERNS`, so `matchFamilyRegex(head)` is unambiguous.

Per #245's bounded-scope rationale, this is a focused follow-up: the synthesized-canonical pattern is generalized to the remaining same-class shape without expanding scope to recorder-vs-strict semantics (#232) or canonical `failure_code` separation (#230).

## Test plan

- [x] `node --test hooks/__tests__/mpl-issue-256-non-runner-canonical.test.mjs` → 14/14 pass
- [x] `node --test hooks/__tests__/*.test.mjs` → 1570/1570 pass (no regression in #244/#245 runner-head suite or any other gate-classify path)
- [x] Each repro from the issue body now resolves to its correct family
- [x] Bare-head and config-file-arg legit shapes still classify correctly
- [x] Unrecognized heads (`biome`, `deno`, `bun` without config) still return `null`

### DIFF
--- a/hooks/__tests__/mpl-issue-256-non-runner-canonical.test.mjs
+++ b/hooks/__tests__/mpl-issue-256-non-runner-canonical.test.mjs
@@ -1,0 +1,176 @@
+/**
+ * #256 — Residual I12 forgery surface for NON-RUNNER gate heads.
+ *
+ * PR #245 fixed the trailing-keyword shape for runner heads
+ * (npm/pnpm/yarn/cargo/go via `classifyRunnerHead`). The structurally
+ * same-class hole remained for the rest of `STRICT_GATE_HEAD_ALLOWLIST`:
+ * heads whose binary name IS the family signal (tsc/eslint/vitest/jest/
+ * mocha/ruff/mypy/playwright/cypress/wdio). Their fallback ran
+ * `matchFamilyRegex(safeCanonical)` against the whole canonical, so a
+ * trailing HARD3 keyword promoted the entry. ANSI-C `$'...'` and
+ * parameter expansion `${...}` wrapped the same forgery into different
+ * quoting forms.
+ *
+ * AC coverage:
+ *   (A) Trailing positional keyword arg → family from the head itself.
+ *       `tsc playwright` → hard1_baseline (not hard3_resilience).
+ *   (B) Quote / expansion wrappers carry no signal.
+ *       `tsc $'playwright'` → hard1_baseline.
+ *       `tsc ${FOO}playwright` → hard1_baseline.
+ *       `eslint $"cypress"` → hard1_baseline.
+ *
+ * The fix synthesizes a canonical of just the head before
+ * `matchFamilyRegex` for the head-as-signal set.
+ */
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+
+import { classifyGateCommand } from '../lib/mpl-gate-classify.mjs';
+
+// ---------------------------------------------------------------------------
+// (A) Trailing keyword arg on non-runner heads
+// ---------------------------------------------------------------------------
+
+test('#256 (A) [data-integrity]: `tsc playwright` classifies as hard1_baseline, not hard3_resilience', () => {
+  assert.equal(classifyGateCommand('tsc playwright'), 'hard1_baseline');
+});
+
+test('#256 (A) [data-integrity]: `eslint cypress` classifies as hard1_baseline, not hard3_resilience', () => {
+  assert.equal(classifyGateCommand('eslint cypress'), 'hard1_baseline');
+});
+
+test('#256 (A) [data-integrity]: `vitest playwright` classifies as hard2_coverage, not hard3_resilience', () => {
+  assert.equal(classifyGateCommand('vitest playwright'), 'hard2_coverage');
+});
+
+test('#256 (A) [data-integrity]: `jest cypress` classifies as hard2_coverage, not hard3_resilience', () => {
+  assert.equal(classifyGateCommand('jest cypress'), 'hard2_coverage');
+});
+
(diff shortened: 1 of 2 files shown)


KyubumShin:
## Scheduled Hermes PR Review

**Scope:** lightweight diff review of PR #261 at `a4064fd`.

### Findings
- ✅ Looks Good — The change correctly narrows the non-runner fallback from `matchFamilyRegex(safeCanonical)` to `matchFamilyRegex(head)` for built-in head-as-family-signal commands, which directly addresses the trailing-keyword / quoting-wrapper forgery class described in #256.
- ✅ Looks Good — Regression coverage is focused and includes representative hard1, hard2, hard3, wrapper quoting, runner-head preservation, and unknown-head rejection cases.
- 💡 Suggestion — Consider deriving or asserting `STRICT_HEAD_AS_FAMILY_SIGNAL` against `STRICT_GATE_HEAD_ALLOWLIST` minus runner heads in a future cleanup, so future built-in head additions cannot accidentally miss this protection. Not blocking for this PR.

### Notes / Caveats
- Reviewed the PR diff only plus nearby classifier context.
- Ran targeted test: `node --test hooks/__tests__/mpl-issue-256-non-runner-canonical.test.mjs` → 14/14 pass.
- Also attempted the broader `node --test hooks/__tests__/*.test.mjs`; it reported 1565/1570 pass with 5 failures outside the changed classifier/test files. I did not treat those as PR-specific blockers in this lightweight review.

---
*Automated periodic review by Hermes Agent.*

PR #229, opened 2026-05-28
Validate no_parallel_explanation content against computed reasons (#214)

KyubumShin:
## Summary

- Closes #214. `schedulerExplanationMissing` now enforces that, for runs where `computed.rejection_reasons` is non-empty, `no_parallel_explanation` MUST reference at least one of those reasons as a canonical token (snake_case, hyphen, or space variant), case-insensitive.
- Tier-id-only explanations (e.g. `'tier 1'` for a tier with `rejection_reasons: ['file_overlap']`) now fail closed.
- Two pre-existing fixtures updated to use canonical reason tokens so the prose explanation aligns with the structured `rejection_reasons` field. Four new #214 tests cover the matrix.

## Why

Codex r17 on PR #213 flagged this: the existing tier-id-mention check stopped `n/a`-style placeholders but let `'tier 1'` slip through with no actual reason named. The summary's structured `rejection_reasons` field already carries the canonical vocabulary — requiring the prose to reuse one of those tokens keeps the two sources of truth in sync.

## Approach

The new check runs after the existing tier-id-mention check. For each `computed.rejection_reasons` entry, the explanation is searched for the exact token, or its hyphen-/space-separated form (case-insensitive). A miss returns `scheduler:no_parallel_explanation_missing_reasons:expected_one_of=[...]` so operators get a concrete fix hint.

Canonical-vocabulary enforcement (rather than stemming / synonym match) is deliberate — the operator can paste the exact `rejection_reasons` value to satisfy the gate; ambiguous paraphrase like "conflicting files" fails so the diagnostic signal stays clear.

## Test plan

- [x] `node --test hooks/__tests__/mpl-require-finalize-artifacts.test.mjs` — 31/31 pass, including 4 new #214 tests.
- [x] Full hook suite — 1267/1267 pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #265, opened 2026-05-30
Close recorder semantic gaps: exit-code masking + I12 round-trip (#232)

### LINKED ISSUE #232: [Exp22 follow-up #220] Recorder semantic gaps — exit code vs leading command + strict/recorder allowlist divergence

KyubumShin:
## Summary

Closes #232. Both surfaces from the issue body are resolved.

## What

**(1) Exit-code masking — `npm test || true` no longer records a passing entry.**

PR #231 stripped the suffix at the classifier so the leading command drove the family classification. The recorder, however, still stored the OVERALL shell exit code from the composite, so a failure that the shell swallowed (`|| true`, `;`, `|`, `&`, newline) would be written as a passing hard2/3 entry. PR #220 deliberately preserved leading-family classification for legitimate consumers (manual gate-evidence writes with redirects, structured-write family probes), so this PR moves the rejection into the RECORDER instead of the classifier:

- `hooks/lib/mpl-gate-classify.mjs` exports a new `compositeRejectReason()` helper. The classifier itself is unchanged — PR #220's contract is preserved.
- `hooks/mpl-gate-recorder.mjs` calls `compositeRejectReason()` before recording any `state.gate_results` entry. Non-null reason → drop the event entirely. `&&` (short-circuits — leading failure propagates) and pure redirects / comments (`>`, `>>`, `&>`, `#`) remain allowed.

**(2) Strict / recorder allowlist divergence — recorder evidence is authoritative.**

The recorder accepts execution-wrapper heads (`docker`, `bash -lc`, `kubectl exec`) that the strict allowlist deliberately rejects. A recorder write of `docker compose run app npm test` re-classified as `null` on the next I12 `STATE_WRITE` check and fired `GATE_COMMAND_FAMILY_MISMATCH` on legitimate recorder evidence.

- `hooks/mpl-gate-recorder.mjs` tags every entry it writes with `source: 'recorder'`. The recorder is authoritative for what it observed at execution time.
- `hooks/lib/mpl-state-invariant.mjs` I12 skips strict re-validation for entries carrying `source: 'recorder'`. Carve-out is narrowly scoped: manual writes that try to pass a wrapper without the marker still get flagged. `mpl-write-guard` already rejects direct edits to `state.gate_results` outside the recorder, so forging the marker would require breaching the broader write surface.

## Why

The issue body acceptance criteria explicitly listed both shapes:

- `npm test || true` does not produce a passing hard2 entry when `npm test` actually failed.
- `docker compose run app npm test` recorder write does not trigger an I12 family mismatch on subsequent `STATE_WRITE` checks (round-trip holds OR I12 treats recorder writes as authoritative).

The chosen approach (recorder-side rejection + provenance marker) preserves the asymmetry called out by codex r6 / r7 on PR #231 — strict and recorder are deliberately different — and adds a per-entry attestation so the asymmetry survives a round trip.

## Test plan

- [x] `node --test hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs` → 20/20 pass
- [x] `node --test hooks/__tests__/*.test.mjs` → 1610/1610 pass (no regression in PR #220 / #229 / #231 suites)
- [x] Issue-body repros covered:
  - `npm test || true` (shell exit 0) → no `gate_results.hard2_coverage` entry
  - `npx playwright test | tee output.log` → no `gate_results.hard3_resilience` entry
  - `npm test ; true` → no `gate_results.hard2_coverage` entry
  - `npm test && echo done` (safe) → entry recorded
  - `npm test > log.txt` (redirect safe) → entry recorded
  - `docker compose run app npm test` recorder write → no I12 mismatch (carries `source: 'recorder'`)
  - Manual wrapper write without `source` marker → I12 still surfaces (carve-out is narrow)

### DIFF
--- a/hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs
+++ b/hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs
@@ -88,6 +88,27 @@ describe('#232 (1) compositeRejectReason detection — unit', () => {
     assert.equal(compositeRejectReason('npm test &> log.txt'), null);
   });
 
+  // Hermes r4 on PR #265: fd duplication (`n>&m`, `n<&m`, `>&-`)
+  // is a redirect operator, not a background-`&`. `2>&1` is the
+  // canonical "merge stderr into stdout" pattern, and the recorder
+  // must NOT drop legitimate logging shapes.
+  it('keeps `2>&1` and other fd-duplication forms (NOT background)', () => {
+    assert.equal(compositeRejectReason('npm test 2>&1'), null);
+    assert.equal(compositeRejectReason('npm test > log.txt 2>&1'), null);
+    assert.equal(compositeRejectReason('npm test 1>&2'), null);
+    assert.equal(compositeRejectReason('npm test 3>&1'), null);
+    assert.equal(compositeRejectReason('npm test >&-'), null);
+    assert.equal(compositeRejectReason('npm test <&3'), null);
+  });
+  it('keeps `bash -lc "npm test 2>&1"` — fd dup inside wrapper payload', () => {
+    assert.equal(compositeRejectReason('bash -lc "npm test 2>&1"'), null);
+  });
+  it('still flags bare `&` background even when redirect operators are also present', () => {
+    // `&` between redirects-and-command is still background.
+    assert.equal(compositeRejectReason('npm test > log.txt &'), 'background');
+    assert.equal(compositeRejectReason('npm test 2>&1 &'), 'background');
+  });
+
   it('keeps trailing comment shapes — `#` is shell-discarded', () => {
     assert.equal(compositeRejectReason('npm test # smoke'), null);
   });
--- a/hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs
+++ b/hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs
@@ -134,6 +134,12 @@ describe('#232 (1) compositeRejectReason detection — unit', () => {
   it('flags `+o errexit -c` style (off-option flag) too', () => {
     assert.equal(compositeRejectReason('bash +o errexit -c "npm test || true"'), 'or_or');
   });
+  it('flags `-O extglob -c` style (shopt option flag with value)', () => {
+    // Hermes r3 on PR #265: bash also exposes shopt via -O / +O,
+    // structurally identical to -o / +o.
+    assert.equal(compositeRejectReason('bash -O extglob -c "npm test || true"'), 'or_or');
+    assert.equal(compositeRejectReason('bash +O nullglob -c "npm test ; true"'), 'semicolon');
+  });
   it('flags `--rcfile <path> -c` (rcfile flag takes value)', () => {
     assert.equal(compositeRejectReason('bash --rcfile /tmp/rc -c "npm test ; true"'), 'semicolon');
   });
--- a/hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs
+++ b/hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs
@@ -97,6 +97,30 @@ describe('#232 (1) compositeRejectReason detection — unit', () => {
     assert.equal(compositeRejectReason('pytest -k "login || logout"'), null);
   });
 
+  // Hermes review on PR #265: shell wrappers (`bash -c`, `sh -c`,
+  // `bash -lc`, ...) hide the payload from the outer quote-aware
+  // scan. The wrapper itself evaluates the masking operator
+  // internally, so the bypass surface must extend into the wrapped
+  // payload.
+  it('flags `bash -lc "npm test || true"` — masking inside shell wrapper payload', () => {
+    assert.equal(compositeRejectReason('bash -lc "npm test || true"'), 'or_or');
+  });
+  it('flags `sh -c "npx playwright test | tee out.log"` — pipe inside wrapped payload', () => {
+    assert.equal(compositeRejectReason('sh -c "npx playwright test | tee out.log"'), 'pipe');
+  });
+  it('flags `bash -c \'npm test ; true\'` — semicolon inside wrapped payload', () => {
+    assert.equal(compositeRejectReason("bash -c 'npm test ; true'"), 'semicolon');
+  });
+  it('flags `/usr/local/bin/bash -lc "npm test || true"` — path-qualified shell head', () => {
+    assert.equal(compositeRejectReason('/usr/local/bin/bash -lc "npm test || true"'), 'or_or');
+  });
+  it('keeps `bash -lc "npm test && echo done"` — `&&` in payload still safe', () => {
+    assert.equal(compositeRejectReason('bash -lc "npm test && echo done"'), null);
+  });
+  it('keeps `bash -lc "npm test"` — no composite in payload', () => {
+    assert.equal(compositeRejectReason('bash -lc "npm test"'), null);
+  });
+
   it('classifier still resolves leading family on composite shapes — PR #220 contract preserved', () => {
     // The classifier is unchanged for these shapes (the recorder is
     // where rejection happens). Asserting here so a future refactor
@@ -321,5 +345,26 @@ describe('#232 (1) recorder integration — masking composites do not produce ga
     const state = postToolUse('docker compose run app npm test', { exit_code: 0 });
     assert.equal(state.gate_results?.hard2_coverage?.source, 'recorder',
       'recorder must tag every entry with source: \'recorder\' so I12 can skip strict re-validation');
+  });
+
+  // Hermes review on PR #265 found this bypass: the masking
+  // operator is hidden inside the shell wrapper's quoted payload, so
+  // the outer compositeRejectReason scan let it through. The
+  // wrapper-payload re-scan now catches it.
+  it('`bash -lc "npm test || true"` (shell-wrapper masking) does NOT produce a hard2_coverage entry', () => {
+    const state = postToolUse('bash -lc "npm test || true"', { exit_code: 0 });
+    assert.equal(state.gate_results?.hard2_coverage, undefined,
+      `recorder must drop wrapper-hidden masking composites, got: ${JSON.stringify(state.gate_results)}`);
+  });
+
+  it('`sh -c "npx playwright test | tee out.log"` (wrapped pipe) does NOT produce a hard3_resilience entry', () => {
+    const state = postToolUse('sh -c "npx playwright test | tee out.log"', { exit_code: 0 });
+    assert.equal(state.gate_results?.hard3_resilience, undefined,
+      `recorder must drop wrapper-hidden pipes, got: ${JSON.stringify(state.gate_results)}`);
+  });
+
+  it('`bash -lc "npm test"` (no payload composite) DOES still record', () => {
+    const state = postToolUse('bash -lc "npm test"', { exit_code: 0 });
+    assert.equal(state.gate_results?.hard2_coverage?.command, 'bash -lc "npm test"');
   });
 });
--- a/hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs
+++ b/hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs
@@ -1,0 +1,325 @@
+/**
+ * #232 — Recorder semantic gaps after PR #231.
+ *
+ * Two structurally-distinct surfaces remained from the strict / recorder
+ * unification work:
+ *
+ *  (1) The recorder cut at the first control operator and classified
+ *      only the leading simple command, BUT it still stored the OVERALL
+ *      shell exit code from the composite. So:
+ *        - `npm test || true`             → leading=npm test (hard2),
+ *                                            shell exit = 0 from `true`
+ *        - `npm test ; true`              → leading=npm test (hard2),
+ *                                            shell exit = 0 from `true`
+ *        - `npx playwright test | tee X`  → leading=npx playwright (hard3),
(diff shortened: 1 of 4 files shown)


KyubumShin:
## Scheduled Hermes PR Review

**Scope:** lightweight diff review of PR #265 at `af8f0de`.

### Findings
- ⚠️ Warning — `compositeRejectReason()` currently treats common stderr/stdout redirection forms as `background` because it flags any bare `&` unless the next char is `&` or `>`. For example, `npm test 2>&1` and `npm test >& log.txt` both return `background`, so `mpl-gate-recorder` will silently drop otherwise-valid gate evidence. The PR explicitly intends to allow redirects because they do not mask exit codes; please handle shell fd-redirection forms such as `2>&1` / `>& file` (or add tests documenting the intended behavior).
- ✅ Looks Good — the recorder-side rejection is the right place to address false PASS rows from `||`, `;`, bare pipe, background, and newline composites while preserving the PR #220 classifier contract.
- ✅ Looks Good — recorder provenance via `source: 'recorder'` cleanly scopes the I12 strict re-validation carve-out to recorder-observed evidence.

### Verification run
- `node --test hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs` → 20/20 pass
- `node --test hooks/__tests__/*.test.mjs` → 1610/1610 pass
- Extra probe: `compositeRejectReason('npm test 2>&1')` → `background`; `compositeRejectReason('npm test >& log.txt')` → `background`

### Notes / Caveats
- Review was diff-focused and lightweight, not a full repository audit.

---
*Automated periodic review by Hermes Agent.*

KyubumShin:
## Scheduled Hermes PR Review

**Scope:** lightweight diff review of PR #265 at `af8f0de`.

### Findings
- ⚠️ **Warning — `hooks/lib/mpl-gate-classify.mjs:786` / `hooks/mpl-gate-recorder.mjs:257`** — `compositeRejectReason()` is quote-aware and intentionally ignores `||`, `;`, and `|` inside quotes, but the recorder classifier still accepts shell wrapper commands such as `bash -lc "npm test"` as legitimate recorder evidence. That leaves an exit-code masking bypass: `bash -lc "npm test || true"` is not rejected, then `classifyRecordedCommand()` still matches `npm test`, and the recorder writes a passing `hard2_coverage` entry with `source: 'recorder'` when the outer shell exit is 0.
  - I reproduced this against this PR branch with a live `mpl-gate-recorder.mjs` invocation; it produced:
    ```json
    "hard2_coverage": {
      "command": "bash -lc \"npm test || true\"",
      "exit_code": 0,
      "source": "recorder"
    }
    ```
  - Suggested fix: either make `compositeRejectReason()` reject masking operators inside shell-wrapper payloads (`bash -lc`, `sh -c`, etc.), or reject shell-wrapper recorder evidence when the wrapped payload contains masking composites. Please add regression coverage for at least `bash -lc "npm test || true"` and ideally `sh -c "npx playwright test | tee out.log"`.

- ✅ **Looks Good** — The narrower non-wrapper cases from the issue are covered well: `npm test || true`, `npm test ; true`, and `npx playwright test | tee output.log` now drop recorder events; `&&` and redirects remain allowed; recorder provenance is tagged; and I12 preserves the manual-write vs recorder asymmetry.

### Notes / Caveats
- Review was diff-focused with targeted context around the changed classifier/recorder/invariant paths.
- Local checks run from a temporary checkout:
  - `node --test hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs` → pass 20/20
  - `node --test hooks/__tests__/*.test.mjs` → pass 1610/1610
- I did not approve/request changes; this is a scheduled `COMMENT`-style review.

---
*Automated periodic review by Hermes Agent.*

KyubumShin:
## Scheduled Hermes PR Review

**Scope:** lightweight diff review of PR #265 at `4caf885`.

### Findings
- ⚠️ **Warning — `hooks/lib/mpl-gate-classify.mjs:828`**: The new shell-wrapper payload extraction still has bypasses for common wrapper forms that include option arguments or nested wrappers. For example, on this PR head:
  - `compositeRejectReason('bash -o pipefail -c "npm test || true"')` returns `null`
  - `compositeRejectReason('bash -lc "bash -c \'npm test || true\'"')` returns `null`
  - and `classifyRecordedCommand(...)` still classifies these as `hard2_coverage`.

  That means the recorder can still write a false passing gate row for a masked failing test when the command is routed through these wrapper shapes. The comment says the wrapper scanner skips flags that take values “conservatively”, but `split(/\s+/)` + the current loop stops at the `pipefail` value before reaching `-c`. The nested-wrapper comment also says one level is sufficient, but the quoted inner payload remains hidden from `scanForMaskingOperator()`.

  Suggested fix: parse the shell `-c` invocation more deliberately (including flags with values such as `-o pipefail`) and either recursively unwrap shell-wrapper payloads or scan nested wrapper payloads after each peel. Please add regression tests for at least `bash -o pipefail -c "npm test || true"` and `bash -lc "bash -c 'npm test || true'"`.

### Notes / Caveats
- Targeted test run: `node --test hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs` passed (`29/29`).
- Broad test run: `node --test hooks/__tests__/*.test.mjs` failed with 5 existing/unrelated-looking failures in this checkout (`1614/1619` pass); I did not attribute those to this PR without deeper investigation.
- Review was lightweight and diff-focused; I checked the changed classifier/recorder paths and minimal local behavior only.

---
*Automated periodic review by Hermes Agent.*

KyubumShin:
## Scheduled Hermes PR Review

**Scope:** lightweight diff review of PR #265 at `1a22cda`.

### Findings
- ⚠️ **Warning — `hooks/lib/mpl-gate-classify.mjs:491`**: `extractShellWrapperPayload()` still misses at least one bash option that takes a separate value: `-O <shopt_option>`. For example, `bash -O extglob -c "npm test || true"` is accepted by bash and exits through the wrapped `|| true`, but the current parser stops at `extglob`, returns `null` from `compositeRejectReason(...)`, and would allow the recorder to record a potentially masked PASS. Suggested fix: add `-O` (and, if this is intended to be a generic bash wrapper parser, audit nearby separate-value shell options) to `SHELL_WRAPPER_FLAGS_WITH_VALUE`, with a regression test mirroring the existing `-o pipefail` case.
- ✅ **Looks Good:** The new `source: 'recorder'` carve-out is narrowly applied to recorder-sourced entries, while manual wrapper writes remain covered by strict I12 re-validation.
- ✅ **Looks Good:** The newly added targeted tests cover the original masking composites, shell-wrapper payloads, nested wrappers, and recorder/I12 round-trip behavior.

### Notes / Caveats
- I ran the targeted test and the broad hook test suite locally from a temporary PR checkout:
  - `node --test hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs` — pass (`35/35`)
  - `node --test hooks/__tests__/*.test.mjs` — pass (`1625/1625`)
- Review was lightweight and diff-focused; I did not perform a full architecture audit beyond the changed files and directly related behavior.

---
*Automated periodic review by Hermes Agent.*

KyubumShin:
## Scheduled Hermes PR Review

**Scope:** lightweight diff review of PR #265 at `e4a76b7`.

### Findings
- ⚠️ **Warning — `hooks/lib/mpl-gate-classify.mjs`: redirect handling false-positive.** `scanForMaskingOperator()` currently treats any bare `&` that is not `&&` or `&>` as `background`. That catches the intended `npm test &` case, but it also rejects common safe fd-duplication redirects such as `npm test > log.txt 2>&1` and `npm test 2>&1`. Those redirects do not mask the command exit code, so the recorder would now silently drop valid gate evidence for a common logging pattern. I verified on this PR head:
  - `compositeRejectReason('npm test > log.txt 2>&1') => background`
  - `compositeRejectReason('npm test 2>&1') => background`
  Consider recognizing `[n]>&[word]` / `[n]<&[word]`-style redirections before classifying `&` as background, and add regression tests for `2>&1` (including inside `bash -lc "..."`).

### Notes / Caveats
- Tests run locally from a temporary checkout:
  - `node --test hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs` → pass 36/36
  - `node --test hooks/__tests__/*.test.mjs` → pass 1626/1626
- This was a lightweight scheduled diff review, focused on the changed files and the new command-classification semantics. It was not a full manual audit of the entire repository.

---
*Automated periodic review by Hermes Agent.*

KyubumShin:
## Scheduled Hermes PR Review

**Scope:** lightweight diff review of PR #265 at `6703df2`.

### Findings
- ✅ Looks Good — The updated `compositeRejectReason()` path now covers the previously noted shell-wrapper bypasses, including `bash -lc "npm test || true"`, nested wrappers, and option/value forms such as `bash -o pipefail -c ...`.
- ✅ Looks Good — The `2>&1` / fd-duplication handling avoids the earlier false positive while still dropping real background forms like `npm test 2>&1 &`.
- ✅ Looks Good — Recorder provenance via `source: 'recorder'` is narrowly handled in I12, and tests preserve the strict/manual-write rejection path for wrapper commands without the marker.
- 💡 Suggestion — `hooks/lib/mpl-gate-classify.mjs` has a small stale/internal-detail comment around `extractShellWrapperPayload()` mentioning an `includes('=')` branch that is not present in the loop. This is non-blocking, but tightening that comment would reduce future reviewer confusion.

### Notes / Caveats
- I reviewed the PR diff and minimal surrounding context only; this was not a full repository design audit.
- Local checks run on a temporary checkout:
  - `node --test hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs` → pass (39/39)
  - `node --test hooks/__tests__/*.test.mjs` → pass (1629/1629)

---
*Automated periodic review by Hermes Agent.*

KyubumShin:
## Scheduled Hermes PR Review

**Scope:** lightweight diff review of PR #265 at `b64f833`.

### Findings
- ⚠️ **Warning — `hooks/lib/mpl-gate-classify.mjs:900`**: shell-wrapper payload extraction still misses masking operators when the wrapper command has a safe trailing redirect/comment after the quoted payload. For example, on this HEAD:
  - `compositeRejectReason('bash -c "npm test || true" 2>&1')` returns `null`
  - `compositeRejectReason('bash -lc "npm test || true" > out.log')` returns `null`
  - `compositeRejectReason('bash -lc "npm test || true" # comment')` returns `null`

  Those shapes are still evaluated by the shell wrapper, so `|| true` can mask the failing gate exit while the recorder may treat the command as legitimate gate evidence. The current `tokens.slice(i).join(' ')` + whole-string quote peel only works when the payload is the final token group. Consider parsing/extracting exactly the `-c` argument (respecting quotes) before trailing redirections/comments, or add targeted tests for wrapper payload plus `2>&1`, `> log`, and `# comment`.

- ✅ **Looks Good:** the direct composite cases (`||`, `;`, pipe, background) and wrapper-hidden composites without trailing suffixes are covered by focused tests, and recorder provenance (`source: 'recorder'`) is narrowly handled in I12.

### Notes / Caveats
- Ran: `node --test hooks/__tests__/mpl-issue-232-recorder-semantic-gaps.test.mjs` — pass (`39/39`).
- Review was lightweight and diff-focused; I did not run the full repository test suite in this scheduled pass.

---
*Automated periodic review by Hermes Agent.*

